PR #2, opened 2014-12-08
Initial work

purplefox:


src/main/java/io/vertx/ext/auth/impl/realms/SimplePrincipalCollection.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  @Override
+  public boolean isEmpty() {
+    return false;
+  }
+
+  @Override
+  public Iterator iterator() {
+    return null;
```
vietj, line 83: shouldn't it return an empty iterator rather ?

purplefox, line 83: The method is never called. Perhaps I should throw UnsupportedOperationExcepttion for clarity...

PR #5, opened 2015-02-19
Docs and javadocs, and a few refinements to the API

purplefox:
@vietj please review :)


src/main/asciidoc/groovy/index.adoc
```diff
@@ -0,0 +1,367 @@
...
+def proxy = AuthService.createEventBusProxy(vertx, "vertx.auth")
+
+// Now do stuff with it:
+
+proxy.login(credentials, { res ->
+})
+
+----
+
+Alternatively you can create an instance of the service directly and just use that locally:
```
vietj, line 65: perhaps we should add a note about using via a proxy versus using directly and explain why one would use one or the other.

vietj:
looks good to me.

PR #9, opened 2015-04-02
Simplifications to API, proper use of principal/credentials, and make Au...

purplefox:
...thProvider first class genned interface

@vietj please review

I have some TODO in the docs. Would like to complete that after milestone4 :)


vietj:
looks good enought to me

PR #12, opened 2015-05-18
Removed AuthService and made AuthProvider/User the main auth interface

purplefox:


purplefox:
@stephane-bastian @pmlopes please review

src/main/asciidoc/groovy/index.adoc
```diff
@@ -1,25 +1,21 @@
...
 currently allows user/role/permission information to be accessed from simple properties files or LDAP servers.
 
+Vert.x auth is also used by vertx-web to handle its authentication and authorisation.
+
 == Basic concepts
 
 _Authentication_ (aka _log in_) means verifying the identity of a user.
 
-_Authorisation_ means verifying a user has the correct access rights for some resource.
+_Authorisation_ means verifying a user is allowed to access some resource.
```
pmlopes, line 18: Not wanting to start a language war here, but should we stick with the British spelling or the American spelling? What I am trying to clear here is that in most of literature you will read: AuthN/AuthZ for "Authentication/Authorization". I understand that Tim is British but for me (no offense) I feel the "Americanized" spelling a bit more convenient...

purplefox, line 18: I'm not american, and actually no-one in the team is American so I don't see why we should use American dialect, let's just use proper English ;)

pmlopes, line 18: I understand I just want to avoid reactions to API users like here:

http://docs.wxwidgets.org/3.0/classwx_colour.html
http://docs.oracle.com/javase/7/docs/api/java/awt/color/package-summary.html

So if you use wxWidgets you refer to "colour" and if you use java you refer to "color". An example more close to us is the JAAS API:

http://docs.oracle.com/javase/7/docs/technotes/guides/security/jaas/JAASRefGuide.html

Where the spelling for obvious reasons is American... Anyway I am fine with either, just wanted to get some reactions on the subject.

purplefox, line 18: Hey, it's not our fault if the JDK and other libraries use the incorrect American spelling ;)

stephanebastian:
There are compile errors on my side, that all seem related to the same root issue => ClusterSerializable.readFromBuffer/writeToBuffer has changed. I am probably not using the latest version from vertx-core. Will update asap

Anyway, glancing through the code, I am wondering if User.isClusterable() is belongs to the API. It seems to me that this is more or less a configuration option or some sort of implementation detail. Maybe this is a parameter that could be specified in the AuthHandler? (=> AuthHandler.isUserClustered(true | false) or smething ? )

Same comment for User AuthProvider.fromBuffer(Buffer buffer). Is this really something that belongs to the API ?

pmlopes:
I agree with Bastian that `User.isClusterable()` does not feel like part of the API since it is a detail about serialization. I think the `ClusterSerializable` interface should be Provider specific implementation. For the rest I am fine since I managed to get token auth working with the new API and i think I can get some OAuth2 prototype implementation too.

src/main/asciidoc/groovy/index.adoc
purplefox, line 18: Which inspired me to suggest this idea: https://twitter.com/timfox/status/600365048847060993 ;)

purplefox:
Stephane: compile errors - you need this PR from core too https://github.com/eclipse/vert.x/pull/1040

User.isClustered - we could put this in the auth handler, but that would prevent us from having some Users which are clusterable and others which aren't with the same handler, but not sure if that is a big issue.

AuthProvider.fromBuffer - this is needed to recreate the User after it has been extracted from a clustered session, i.e. so when a request arrives on another node of the cluster we still know the user is logged in, and has various cached permissions and roles. Only the AuthProvider knows how to recreate the User so I'm not sure how we could put this anywhere else. Any ideas?

purplefox:
Folks I am going to merge this as time is of the essence here. Will address the above two points afterwards.

stephanebastian:
Right, time is definitely a key concern right now
Quick question:
An AuthProvider returns a User. Then it is saved to a Session (clustered or not) right? 
Then why is the AuthProvider the one who knows how to recreate the User? I would think that the logic required to return a User maybe specific to AuthProvider but in any case when the user is returned, the only thing we are interested in is the User interface (roles / permissions). So I would assume that 'recreating' a User from a Buffer is no longer related or specific to an AuthProvider. Am I missing something?

purplefox:
The User must know about the AuthProvider otherwise if subsequent calls to hasRole/hasPermission are made after it has been deserialized how is it going to know which AuthProvider actualls handles those operations?

stephanebastian:
Extremely hectic today, sorry for the late reply,

Didn't get a chance to get rid of the compile error yet... So my understanding is that the Auth module stores the User in a session between requests. It doesn't only store the userId but the whole user gets serialized in session. right?
Lets quickly look at the following flow:
- A user enters its username/password. The AuthProvider does its thing and returns a User.
- A cookie (right now a session cookie I guess) is sent back to the browser so we know who the user is, when he comes back
- The user request a different page, we get the cookie and somehow recreate the user (well.. right now using the Session) --> This logic is hardcoded in the Auth module. 
  IMHO this code should not be hardcoded but rather extracted in an AuthProvider. To me this is a SessionAuthProvider or something along these lines. It validates and recreates a User from the Session cookie. 
  The JWTAuthProvider will do something very similar using a Cookie but no session

If we do not hard-code the session thing in the Auth module, this is way cool because we can plugin any logic we want, plus the User and AuthProvider interfaces are reduced to their bare minimum.

Starting tomorrow, I've got 3 days to work on Vertx. Do you want me to implement this and submit a PR if I can make it work this way?

purplefox:
Stephane - I'm not sure what you mean by "Auth Module". Can you clarify please?

stephanebastian:
the Vertx-Auth module sorry

stephanebastian:
BTW I forgot to thank you for this great PR. Thx a lot

purplefox:
I am confused and don't understand your recent (large) comment. Perhaps it is the terminology you are using (?)

In particular "This logic is hardcoded in the Auth module. IMHO this code should not be hardcoded but rather extracted in an AuthProvider"

Perhaps you could you rephrase it, as right now I don't understand that point you are making. Sorry :(

purplefox:
I thought about this some more, and I think you might be suggesting that we create different AuthProvider instances that take a buffer in the call to authenticate (?)

But I'm not really sure how this work in practice. If you could elaborate that would be great :)

stephanebastian:
Hey Tim, 
I wrote that comment in a hurry while working on another project for a customer, attending a conf call and replying to dozens of emails so it's not surprising that my post was ... confusing ;) multi-tasking is not always a good thing...

If I've got to summarize things, I guess what I am saying is that authenticating a user from a Session could be done in a SessionAuthProvider. In the same vein, authenticating a user from JWT will be done in a JWTAuthProvider based on the Paulo's code.
Right now we don't have a SessionAuthProvider, instead we expect subsequent requests to be authenticated via a sessionId. This is what I meant by hardcoded, the flow is harcoded.
It would be great to authenticate subsequent requests via JWT for instance (not using a sessionid, session-less). 

IMO, we've got to change very little code for this to work. I will work on the details tomorrow and submit a PR along with examples/tests. Working code is always easier to understand than emails..

purplefox:
In the case of a clusterable User, right now we only authenticate once, then we store the User in the session. We don't authenticate via a session ID - we just deserialize the user from a session. deserialization != authentication.

It seems to me that having a SessionAuthProvider would make things more complex - as it would mean authentication would be done every time a request hits a new node - and you'd actually need two AuthProvider instances - one to do the actual authentication and one to do the session part.

PR #13, opened 2015-05-19
JWT Auth provider implementation

pmlopes:
Signed-off-by: Paulo Lopes pmlopes@gmail.com

This is a full JWT/JWS implementation in par with http://jwt.io/

| Feature | Implemented | Comments |
| :-- | :-: | :-- |
| Sign | ✔ |  |
| Verify | ✔ |  |

| Check | Implemented | Comments |
| :-- | :-: | :-- |
| iss | ✔ |  |
| sub | ✔ |  |
| aud | ✔ |  |
| exp | ✔ |  |
| nbf | ✔ |  |
| iat | ✔ |  |
| jti |  | unique ids are not implemented |

| Algorithm | Implemented | Comments |
| :-- | :-: | :-- |
| HS256 | ✔ |  |
| HS384 | ✔ |  |
| HS512 | ✔ |  |
| RS256 | ✔ |  |
| RS384 | ✔ |  |
| RS512 | ✔ |  |
| ES256 | ✔ | Elliptic Curve might require Java Cryptography Extension (JCE) Unlimited Strength Jurisdiction Policy |
| ES384 | ✔ | Elliptic Curve might require Java Cryptography Extension (JCE) Unlimited Strength Jurisdiction Policy |
| ES512 | ✔ | Elliptic Curve might require Java Cryptography Extension (JCE) Unlimited Strength Jurisdiction Policy |


stephanebastian:
Hey Paulo, 
two words: You ROCK! ;)

pmlopes:
i am now making a `vertx-web` handler and a demo app

stephanebastian:
Great!
On a different note, I left a comment on the PR that Tim commited earlier today: 
https://github.com/vert-x3/vertx-auth/pull/12

Tomorow I will see if we can get rid of the AuthProvider.isClusterable and FromBuffer. I will also hack an OAuthAuthProvider (Facebook, Twitter or something).
If all goes well, the plan is to hack another AuthProvider where we can chain multiple providers so that you can log in using Facebook or Twitter or JDBC or xyz in the same web app
Will keep you guys posted

src/main/java/examples/Examples.java
```diff
@@ -79,6 +80,15 @@ public void example4(Vertx vertx, Realm realm) {
...
+
+    JsonObject config = new JsonObject()
+        .put("keyStoreURI", "classpath:///keystore.jceks")
+        .put("keyStoreType", "jceks")
+        .put("keyStorePassword", "secret");
+
+    AuthProvider provider = JWTAuth.create(vertx, config);
+  }
 //
 //  public void example0_1(Vertx vertx, JsonObject config) {
```
purplefox, line 112: I think an example of authenticating using a jwt token would be useful

src/main/java/io/vertx/ext/auth/jwt/JWTAuth.java
```diff
@@ -0,0 +1,41 @@
...
+ * @author Paulo Lopes
+ */
+@VertxGen
+public interface JWTAuth extends AuthProvider {
+  static JWTAuth create(Vertx vertx, JsonObject config) {
+    return new JWTAuthProviderImpl(vertx, config);
+  }
+
+  @Fluent
+  JWTAuth generateToken(JsonObject payload, JWTOptions options, Handler<AsyncResult<String>> resultHandler);
```
purplefox: Does generateToken need to be async? Is it likely to block/take a long time to complete?

src/main/java/io/vertx/ext/auth/jwt/impl/JWT.java
```diff
@@ -0,0 +1,210 @@
...
+import java.util.*;
+
+/**
+ * JWT and JWS implementation draft-ietf-oauth-json-web-token-32.
+ *
+ * @author Paulo Lopes
+ */
+public final class JWT {
+
+  private static final Charset UTF8 = Charset.forName("UTF-8");
```
purplefox: Can use StandardCharsets.UTF_8

src/main/java/io/vertx/ext/auth/jwt/impl/JWTAuthProviderImpl.java
```diff
@@ -0,0 +1,173 @@
...
+
+  private static final JsonObject EMPTY_OBJECT = new JsonObject();
+  private static final JsonArray EMPTY_ARRAY = new JsonArray();
+
+  private final JWT jwt;
+
+  private final String rolesClaimKey;
+  private final String permissionsClaimKey;
+
+  public JWTAuthProviderImpl(Vertx vertx, JsonObject config) {
```
purplefox: vertx parameter is not used

src/main/java/io/vertx/ext/auth/package-info.java
```diff
@@ -186,6 +186,42 @@
...
+ *
+ * To create an instance of the provider you use {@link io.vertx.ext.auth.jwt.JWTAuth}. You specify the configuration
+ * in a JSON object.
+ *
+ * Here's an example of creating a JWT auth provider:
+ *
+ * [source,java]
+ * ----
+ * {@link examples.Examples#example5}
+ * ----
```
purplefox, line 220: I think it would be useful to also have an example of generating a token, and of authenticating using a token.

src/main/java/io/vertx/ext/auth/package-info.java
```diff
@@ -186,6 +186,42 @@
  * `ldap-system-username`:: TODO
  * `ldap-system-password`:: TODO
  *
+ * == The JWT implementation
+ *
```
purplefox, line 190: Perhaps say a few sentences about what this auth provider is used for -i.e. what kinds of use cases

purplefox:
Thanks Paulo.

A few comments above, but on the whole, looks great :)

pmlopes:
I'll try to fix this tomorrow, cheers!

pmlopes:
all changes in, please review

purplefox:
Hi Stephane,

The approach we take in most of Vert.x is to use the upper case for acronyms, e.g. JDBCClient, SQLClient, not JdbcConnection, SqlClient, so I think how Paulo has done it is consistent with that.

purplefox:
Regarding discussion about issues with the auth API I am available all day today on IRC, or we can discuss on the google group. I don't mind which  :)

stephanebastian:
Paulo,
Been rethinking about some minor aspects of this PR. I am wondering if most (all? ) of the code that is in JWTAuthProviderImpl.authenticate (the code that validates a JWT token) should be moved outside the AuthProvider, maybe inside JWT? 
After all it would probably be valuable to validate a JWT outside an AuthProvider.
What do you think?

pmlopes:
Stephane, i don't really agree. JWT class is only concerned with sign and verify of tokens. This is the basis of jwt, sign a json object and verify is a signature is valid.

Then the provider builds on top of this by verifying all _optional_ claims that are usefull for authorisation.

For example an empty json can be used as a token and in this case your _valid_ principal is {} and no checks from auth provider are asserted.

This previous situation can lead to replay/collision attacks since empty principals always have the same signature, in order to fix this i need to implement the optional claim jti, using something like a cluster wide counter.

stephanebastian:
Yep, agree that JWT is only concerned about signing and verifying.
I though methods to check common claims would have been useful outside JWTAuthProvider:
- checkExpirationTime(long now)
- checkIssuedAt(long now)
- checkNotBefore(long now)
  ...

PR #14, opened 2015-05-20
Jdbc auth

purplefox:


stephanebastian:
I had not seen the bug, but yes there is also a relationship between Role and Permission. 

In the end you've got the following table:
A Role table: simply list all roles
A Role_Permission table: list all permissions per role
A User_Role table: list all roles assigned to users.

To get the list of permission for a user, you get all roles assigned to the user then you get all permissions for each role and you are good

stephanebastian:
This PR is great. Does it mean that you plan to remove the dependency on Shiro? The only thing stopping us from doing it is the LDAP provider. 
Do you plan to move the LDAP provider to its own project vertx-auth-ldap?. 
Vertx-Auth would be clean from external deps.Sweet!

purplefox:
I think we should separate out the different impls into Maven submodules as @pmlopes has suggested.

Let's do this after the JWT and JDBC PRs are merged

stephanebastian:
I read @pmlopes post, I am glad to see it will make it before 3.0 gets released

purplefox:
@pmlopes could you also review this? If it looks ok I will merge it in.

purplefox:
I have added a bunch more changes to this PR as its quicker that way than submitting another one and waiting for review and we need to move fast now.

The latest changes remove to the fromBuffer method from AuthProvider and various other related changes.

purplefox:
I could be wrong, but I don't see how knowing that SHA-512 hashing is used will enable you to crack the database.I believe the hashes are one way, and the salt means dictionary attacks won't work (?)

pmlopes:
True, it is late...
-----Original Message-----

pmlopes:
I gave it some more though and the PR is good, but i am also thinking why do we need roles and permissions? why the distinction since from the API there is no relation between them. Why not just a single "GrantedAuthority" like on Spring security? roles can be derived from the authorities if one needs hierarchy otherwise it ends up being a simple model.

http://docs.spring.io/spring-security/site/docs/current/apidocs/org/springframework/security/core/GrantedAuthority.html

http://docs.spring.io/spring-security/site/docs/current/apidocs/org/springframework/security/core/package-summary.html

purplefox:
Paulo-

Can you explain GrantedAuthority to me? From the javadoc it is not obvious to me what it represents

stephanebastian:
@purplefox Thx for removing fromBuffer(Buffer buffer) and isClusterable(). I could have lived with it but it's better without ;)

stephanebastian:
Guys another quick comment. 
hasRole() and hasPermission() are async. Do we _really_, really, really need this ;)? It makes checking permission / roles cumbersome (nested handlers hell...) 

Yesterday I had a chat with Tim and we talked about some providers not being able to return all roles and permissions. Are we sure that this is the case ? To be honest I never came across such use cases. Usually roles and permisions are stored in a DB, LDAP, noSql store. etc.. and you can always get the roles / permission of a user

Having said that, if some providers effectively can't return all roles / permissions then we've got to keep hasRole/hasPermission async.
Does someone have an example of such providers?

Otherwise, we can safely change the signature of hasRole / hasPermission to be

```
User {
    boolean hasRole(String role); 
    boolean hasPermission(String permission)
}
```

stephanebastian:
@purplefox 
Just to be accurate, I am _not_ asking to add getRoles() and getPermission(), I'll do it in a subclass. just asking if hasRole/hasPermision really, really, really have to be async :)

pmlopes:
Basicaly spring has: principals and authorities. A principal is a user and a authority is what he is granted to do. Now we can make spring work with roles if we create an authority: "admin:read" or just with authorities by creating a authority "read". 

-----Original Message-----

purplefox:
So basically authority is just a string and it can mean whatever you want it to mean..

purplefox:
I don't think we can assume it's possible to return all roles / permissions in the general case.

For example, an implementation of hasRole could look something like this:

if (role.startsWith("public.")) {
hasRole = true;
}

So the set of roles that the user has is not really well defined.

purplefox:
Ok, that's an interesting suggestion (using an authority approach).

I also think we are now talking about several things on this PR, including Stephane's other suggestions.

So I suggest that we merge this PR as is, and I also need to merge it into the JWT auth provider...

Then can you both create issues for your suggestions @pmlopes: authorities and @stephane-bastian : async or sync permissions/roles checks and we can continue the discussion there.

Agreed?

stephanebastian:
ok let's do that

Issue #16, opened 2015-05-26
Refactor roles/permissions to authorities


pmlopes:
Currently we have `user.hasRole(String)` and `user.hasPermission(String)` even though the separation of `roles` and `permissions` sounds logic there are some deficiencies with the current implementation.

The first false assumption is that roles and permissions are independent and relate to a user. So say that user `Paulo` has role `dev` and permission `push`. Now lets assume the whole application knows about the roles `dev` and `devops` and a `dev` role can `push` code and a `devops` role can `push` releases into production.

Now with our model if we want to assert if `Paulo` can `push` to production we need to do:

```
paulo.hasRole("devops") && paulo.hasPermission("push")
```

The logic here would fail since `false` && `true` is `false`. Now just say that because I am the lead developer i might have `read` access to prod to pull logs or something... so now `Paulo` is a `dev` with `push` permissions and `devops` with `read` permissions.

The same check from above will assert `true` even tough it is not correct. The `hasPermission` has no context about from which `role` to check the permission for.

How can we fix this:

We need to refactor the `User` to have a method `hasAuthority(String)` if we like the Spring style or `isPermitted(String)` if we like the Shiro style. After reading both I am tending for Shiro...

Now the assertion is a String, both Spring and Shiro have conventions about the format of the Strings. Looking at Shiro a authority has the format:

```
<role or domain>[:<action>[:<instance>]]
```

examples would be:

```
printers:print:color_printer_1st_floor
devops:read
dev:push
```

The impact is that using this approach we can have the same model fitting all current implementations:
- shiro
- jwt
- jdbc

The Strings are opaque so they only have specific meaning to the implementations and no assumptions on the format is made by `vertx-auth`, the usage of the format above is just an example and if one prefers can use String notation:

```
Role_devops
Role_dev
devops:push
devops:read
dev:push
```

Since the assumption is that the String are opaque it is to the implementation of `auth` providers to define if they support or not wildcards, e.g.: `printers:print:*`.

stephanebastian:
Hey Paulo,
At the end of the day, an AuthProvider will read roles and permissions from a DB, an LDAP server, etc..
Roles and Permissions may be structured as above (domain:action:instance), support wildcards or not. 
The real issue here is to provide the _correct_ implementation for hasPermission() and hasRole() => Do we need to support wildcards, do we need to support domain:action:instance, etc. ? 
The thing is that we don't know in advance because this is something that is application specific. 
[...]

PR #17, opened 2015-05-27
Make documentation explicit on the relation between roles and permissions

pmlopes:
This is a sum up of the discussions on IRC and extra examples for matching with wildcards

Relates to #16.

### DIFF
--- a/vertx-auth-common/src/main/java/io/vertx/ext/auth/package-info.java
+++ b/vertx-auth-common/src/main/java/io/vertx/ext/auth/package-info.java
@@ -85,8 +85,8 @@
  *
  * To check if a user has a specific role you use {@link io.vertx.ext.auth.User#hasRole},
  * to check if a user has all the specified roles you use {@link io.vertx.ext.auth.User#hasRoles},
- * to check if a user has a specific permission you use {@link io.vertx.ext.auth.User#hasPermission},
- * to check if a user has all the specified permissions you use {@link io.vertx.ext.auth.User#hasPermissions}.
+ * to check if a user has a specific permission you use {@link io.vertx.ext.auth.User#isPermitted},
+ * to check if a user has all the specified permissions you use {@link io.vertx.ext.auth.User#isPermitted}.
  *
  * The results of all the above are provided asynchronously in the handler.
  *
--- a/vertx-auth-common/src/main/java/io/vertx/ext/auth/package-info.java
+++ b/vertx-auth-common/src/main/java/io/vertx/ext/auth/package-info.java
@@ -74,8 +74,7 @@
  *
  * Once you have an {@link io.vertx.ext.auth.User} instance you can call methods on it to authorise it.
  *
- * to check if a user has a specific permission you use {@link io.vertx.ext.auth.User#isPermitted},
- * to check if a user has all the specified permissions you use {@link io.vertx.ext.auth.User#isPermitted}.
+ * to check if a user has a specific permission you use {@link io.vertx.ext.auth.User#isPermitted}.
  *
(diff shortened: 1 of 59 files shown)


pmlopes:
With the 3 commits the refactor is complete.

vertx-auth-common/src/main/java/io/vertx/ext/auth/package-info.java
```diff
@@ -72,10 +74,8 @@
  *
  * Once you have an {@link io.vertx.ext.auth.User} instance you can call methods on it to authorise it.
  *
- * To check if a user has a specific role you use {@link io.vertx.ext.auth.User#hasRole},
- * to check if a user has all the specified roles you use {@link io.vertx.ext.auth.User#hasRoles},
- * to check if a user has a specific permission you use {@link io.vertx.ext.auth.User#hasPermission},
- * to check if a user has all the specified permissions you use {@link io.vertx.ext.auth.User#hasPermissions}.
+ * to check if a user has a specific permission you use {@link io.vertx.ext.auth.User#isPermitted},
+ * to check if a user has all the specified permissions you use {@link io.vertx.ext.auth.User#isPermitted}.
```
purplefox: this should be removed

vertx-auth-common/src/main/java/io/vertx/ext/auth/package-info.java
```diff
@@ -86,10 +86,13 @@
  * {@link examples.Examples#example2}
  * ----
  *
- * === Caching roles and permissions
+ * Matching of Permissions is expected to be done using String equality unless otherwise stated by the underlying
+ * implementation.
```
purplefox: We should probably say something that the actual meaning of the permissions is up to the provider. I don't think strict equality is required.

vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/JDBCAuth.java
```diff
@@ -33,17 +33,12 @@
...
   /**
-   * The default query to retrieve all roles for the user
+   * The default query to retrieve all permissions for the user
    */
-  static final String DEFAULT_ROLES_QUERY = "SELECT ROLE FROM USER_ROLES WHERE USERNAME = ?";
-
-  /**
-   * The default query to retrieve all permissions for the role
-   */
-  static final String DEFAULT_PERMISSIONS_QUERY = "SELECT PERM FROM ROLES_PERMS RP, USER_ROLES UR WHERE UR.USERNAME = ? AND UR.ROLE = RP.ROLE";
```
purplefox, line 46: Shouldn't we preserve the previous functionality and use a prefix "role:" to signify role. I think in practice most users using the JDBC auth will be using a user/role/permission model.

vertx-auth-shiro/src/test/java/io/vertx/ext/auth/test/shiro/PropertiesShiroAuthProviderTest.java
```diff
@@ -38,4 +39,31 @@ protected JsonObject getConfig() {
     return config;
   }
 
+  @Test
```
purplefox, line 42: I think we still need the roles tests...

pmlopes:
docs and support for roles is restored on JDBC still need to document this and make the prefix a config option.

pmlopes:
So current PR is now complete i think. We have streamlined the API with just: `isPermitted` and for providers that support the concept of roles there is prefix `role:` which is configurable we can run assertions like: `isPermitted("role:admin")`.

The role support is implemented for:
- jdbc
- shiro

The docs have been updated to reflect this change.

PR #18, opened 2015-06-02
Mongo AuthProvider

remi128:
Another implementation of auth for mongo. Could anybody please review that BEFORE the next api-refactoring? :-) 
The system supports administration of crypt behaviour for user passwords. Possible solutions are
NO_SALT - passwords aren't crypted
COLUMN - passwords are crypted, the salt is saved inside the user record itself
EXTERNAL - passwords are crypted by using a value which is stored somewhere outside the application
The definition of the encryption is done by the interface HashStrategy, which should be possible to be used global. For the encryption itself i stole the methods from Tim's JDBC-Auth.
For all three versions test classes are existing, where one can see the use of the auth provider

For the user handling i added an interface UserFactory, which is used to create instances of User. The default implementation is creating a simple instance of a user with username, password etc. 
Inside one of my real projects i am implementing a UserFactory which creates a Customer with more information, based on the User interface.

The next step i will add a package description, hoping that the api will stay robust for a while.


vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/HashStrategy.java
```diff
@@ -0,0 +1,93 @@
...
+   *          the unhashed password
+   * @param user
+   *          the user to get the salt for. This paramter is needed, if the {@link SaltStyle#COLUMN} is declared to be
+   *          used
+   * @return the hashed password
+   */
+  String computeHash(String password, User user);
+
+  /**
+   * Retrieve the password from the user, or as clear text or as hashed version, depending on the definition
```
sfitts, line 57: nit: `or as clear test or as hashed...`  -- extra or at the begining

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/HashStrategy.java
```diff
@@ -0,0 +1,93 @@
...
+   * 
+   * @param user
+   *          the user to get the stored password for
+   * @return the password or as hashed version or as cleartext, depending on the preferences
+   */
+  String getStoredPwd(User user);
```
sfitts, line 63: To be honest I'm having a hard time understanding the need for this method.  The notion that I might be able to retrieve the cleartext password seems... undesirable.  I see the use of this as part of the password checking in MongoAuthImpl, but I guess I'd prefer an API that encapsulated this in a password check method.  So something like `checkPassword(User user, String password)`.  I know this is a largely stylistic concern, but it is driven by the fact that I would like to use this and would likely replace the default hashing with BCrypt and the current API makes that difficult.

vertx-auth-mongo/addUpstream.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
sfitts: Vertx newbie question -- is it common to include this kind of script?

remi128: vert.x newbie reply: i have no idea :-)
it was just a helper for myself inside my fork

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/HashStrategy.java
```diff
@@ -0,0 +1,93 @@
...
+  String computeHash(String password, User user);
+
+  /**
+   * Retrieve the password from the user, or as clear text or as hashed version, depending on the definition
+   * 
+   * @param user
+   *          the user to get the stored password for
+   * @return the password, either as hashed version or as cleartext, depending on the preferences
+   */
+  String getStoredPwd(User user);
```
sfitts, line 63: To be honest I'm having a hard time understanding the need for this method. The notion that I might be able to retrieve the cleartext password seems... undesirable. I see the use of this as part of the password checking in MongoAuthImpl, but I guess I'd prefer an API that encapsulated this in a password check method. So something like `checkPassword(User user, String password)`. I know this is a largely stylistic concern, but it is driven by the fact that I would like to use this and would likely replace the default hashing with BCrypt and the current API makes that difficult.

vertx-auth-mongo/addUpstream.sh
pmlopes: This is pure git usage it should not be part of the PR since once it gets merged the script has no usage

vertx-auth-mongo/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>vertx-auth-mongo</artifactId>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-auth-common</artifactId>
+    </dependency>
+	<dependency>
```
pmlopes: Please fix the indentation

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
```diff
@@ -0,0 +1,407 @@
...
+  }
+
+  @Override
+  public void authenticate(JsonObject authInfo, Handler<AsyncResult<User>> resultHandler) {
+    String username = authInfo.getString(this.usernameCredentialField);
+    String password = authInfo.getString(this.passwordCredentialField);
+
+    // Null username is invalid
+    if (username == null) {
+      resultHandler.handle((Future.failedFuture(new AuthenticationException("Username must be set."))));
```
pmlopes: what is the advantage of creating an exception here instead of just passing an error message string?

usually creating exceptions incur on a jvm performance penalty while unwinding the stack...

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
```diff
@@ -0,0 +1,407 @@
...
+      return;
+    }
+    if (password == null) {
+      resultHandler.handle((Future.failedFuture(new AuthenticationException("Password must be set."))));
```
pmlopes: what is the advantage of creating an exception here instead of just passing an error message string?

usually creating exceptions incur on a jvm performance penalty while unwinding the stack...

vertx-auth-mongo/addUpstream.sh
remi128: ok

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/HashStrategy.java
pmlopes, line 63: The HashStrategy is a mongo implementation detail interface. The common API does not force it. What you are asking is to have a way to change the the hash algorithm.

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/HashStrategy.java
```diff
@@ -0,0 +1,93 @@
...
+ * @author <a href="http://tfox.org">Tim Fox</a>
+ */
+public interface HashStrategy {
+
+  /**
+   * Password hash salt configuration.
+   * <ul>
+   * <li>NO_SALT - password hashes are not salted.</li>
+   * <li>COLUMN - salt is in a separate column for each user in the database.</li>
+   * <li>EXTERNAL - salt is NOT stored in the database, but defined as external value like application preferences or so
```
pmlopes, line 36: just out of curiosity, if you have EXTERNAL why not just use a keystore file? this way you can store the algorithms you like in a secure way and also solve the question bellow about using other algorithms.

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
```diff
@@ -0,0 +1,407 @@
...
+    }
+    AuthToken token = new AuthToken(username, password);
+
+    JsonObject query = createQuery(username);
+    mongoService.find(this.collectionName, query, res -> {
+
+      try {
+        if (res.succeeded()) {
+          User user = handleSelection(res, token);
+          vertx.getOrCreateContext().put(CURRENT_PRINCIPAL_PROPERTY, user);
```
pmlopes: why is this needed?

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/HashStrategy.java
sfitts, line 63: Would that be done here or via the common API?  My reading of the code was that I would do that here by creating a replacement for `DefaultHashStrategy` (since that's where the actual password hashing occurs).  I'd like to use the hashing implementation provided by BCrypt instead of that code.  Am I reading this wrong?

remi128, line 63: Actually this method is obsolet, cause it is returning simply the password from the user. I overtook and renamed that method from JDBCHashStrategy, where it is running a bit different. The hashing is done by method computeHash(String password, User user)

remi128, line 63: p.s.: inside this method the decision is done, wether the password is crypted or not - it depends on the defined saltstyle. The unhashed version is still inside, cause i made the experience that during development time it is simply timesaving. IMO even the column based crypting is not secure enough.
BCrypt looks interesting. Do you know wether jBCrypt is nonblocking?

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
remi128: i'll change

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
remi128: i'll change

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/HashStrategy.java
remi128, line 36: yes, good idea. I'll check that later

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
remi128: old burden, will be out

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/HashStrategy.java
sfitts, line 63: Thanks for the update.  I thought I saw it being used in `MongoAuthImpl.examinePassword` at line 167.  Is that code not active either (looks to be, but I have to admit that I haven't debugged through this yet).

BCrypt doesn't block and doesn't introduce any contention (that I've seen), but it is CPU intensive (intentionally so).  I use it for password hashing in all my projects as it is solid, easy to use, and secure.

BTW -- I agree that any scheme where you have to store the salt is undesirable, so personally I'd never use the column based hashing.

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
```diff
@@ -0,0 +1,406 @@
...
+
+  /**
+   * Examine the given user object. Returns true, if object fits the given authentication
+   * 
+   * @param userObject
+   * @param authToken
+   * @return
+   */
+  private boolean examinePassword(User user, AuthToken authToken) {
+    String storedPassword = getHashStrategy().getStoredPwd(user);
```
sfitts, line 191: FYI -- here is the use of `getStoredPwd` that I was referencing above.

remi128, line 191: Yes, this is true. But what i meant is, that the method getStoredPassword is returning simply the password from the user like it was created. 
You may want to take a look to the method MongoUserFactory.createUser(String username, String password, List<String> roles, List<String> permissions, AuthProvider authProvider) 
Here a new user is created and the password is modified following the SaltStyle of the AuthProvider. Thus: here the password is crypted or not - depending on the SaltStyle.

Later on the method getStoredPassword only returns this result - crypted or not - and nothing else. Because of that i was writing, that this method is normally obsolet, cause it has no own functionality. I think i will drop it.

remi128:
Hey,

i just added the documentation within the package-info. When i am performing mvn install locally, the result of the documentation is not like i would expect that. Could somebody check the package-info please?

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
```diff
@@ -0,0 +1,324 @@
...
+ * 
+ * @author mremme
+ */
+
+public interface MongoAuth extends AuthProvider {
+
+  /**
+   * The property name to be used to set the name of the collection inside the config
+   */
+  public static final String PROPERTY_COLLECTION_NAME           = "collectionName";
```
purplefox: No need for "public static final". If you declare it as a String in a Java 8 linterface it will be implicitly public static final

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
```diff
@@ -0,0 +1,324 @@
...
+import io.vertx.ext.auth.mongo.impl.MongoUserFactory;
+import io.vertx.ext.mongo.MongoClient;
+
+/**
+ * An extension of AuthProvider which is using {@link MongoClient} as store
+ * 
+ * @author mremme
+ */
+
+public interface MongoAuth extends AuthProvider {
```
purplefox, line 38: Should be annotated with @VertxGen

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
```diff
@@ -0,0 +1,324 @@
...
+    return new MongoAuthImpl(vertx, mongoClient, config, userFactory);
+  }
+
+  /**
+   * Set the name of the collection to be used. Defaults to DEFAULT_COLLECTION_NAME
+   * 
+   * @param collectionName
+   * @return
+   */
+  public MongoAuth setCollectionName(String collectionName);
```
purplefox, line 163: Should be annotated with @Fluent (and others that also return a reference to this)

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
```diff
@@ -0,0 +1,324 @@
...
+
+  /**
+   * Creates an instance of MongoAuth by using a {@link MongoUserFactory}
+   * 
+   * @param vertx
+   * @param serviceName
+   * @param config
+   * @return
+   */
+  public static MongoAuth create(Vertx vertx, String serviceName, JsonObject config) {
```
purplefox: I understand why you have added this method, but I think it should be sufficient to just provide create methods which take a MongoClient. If the user wants to use it with an event bus proxy they can just create the proxy themselves and pass it in.

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/UserFactory.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * UserFactory is used by an AuthProvider to create an instance of {@link User}. By using this interface one can use own
+ * implementations of {@link User} without changing the {@link AuthProvider} itself. Additionally the factory methods
+ * should provide needed automatisms in the creation of a user
+ * 
+ * @author mremme
+ */
+
+public interface UserFactory {
```
purplefox: I'm not sure I understand the value of this interface and why it is exposed to the developer.

Can you give a concrete example of why a developer might use this interface?

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoUser.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+
+  /**
+   * Check a series of roles
+   * 
+   * @param roles
+   * @param resultHandler
+   */
+  protected void doHasRoles(Set<String> roles, Handler<AsyncResult<Boolean>> resultHandler) {
```
purplefox: This method doesn't seem to be used anywhere

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoUser.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+
+  /**
+   * Check a series of permissions
+   * 
+   * @param permissions
+   * @param resultHandler
+   */
+  protected void doHasPermissions(Set<String> permissions, Handler<AsyncResult<Boolean>> resultHandler) {
```
purplefox: This method is also not used anywhere

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
```diff
@@ -0,0 +1,406 @@
...
+ * An implementation of {@link MongoAuth}
+ * 
+ * @author mremme
+ */
+public class MongoAuthImpl implements MongoAuth {
+  private static final Logger log                     = LoggerFactory.getLogger(MongoAuthImpl.class);
+
+  private final Vertx         vertx;
+  private MongoClient         mongoClient;
+  private String              usernameField           = DEFAULT_USERNAME_FIELD;
```
purplefox: We don't align columns in the Vert.x codestyle

purplefox:
Hi @remi128  looks good and thanks for refactoring this for the new API :)

Just a few comments (above)

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
remi128: ok

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
remi128, line 38: ok

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
remi128, line 163: ok

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
remi128: you're right, done

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/UserFactory.java
remi128: Currently i am working on an online project, where i am using vertx. Its my first project, where i want to check, wether we will perform a technology switch with my firm from our NetRelay to vertx. 
Inside the project a user must authenticate, so that i am using the mongo-auth, of course. But the interface of the User from vertx is defining only a minimum of the information and structure, like i a need inside my project. So i could use the getPrincipal() method and use the json object to create my needed instance from out of that. Or i could define my UserFactory to create the needed instance as soon. 
The second argument in my opinion is, that by using the UserFactory to create new instances of user, i can ensure that the convention of field names is kept.

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoUser.java
remi128: yes, will remove

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
remi128: This i don't understand, my english seems not to be enough :-)

purplefox: In vert.x we do this:

private String usernameField = DEFAULT_USERNAME_FIELD;

not

private String    <tab><tab><tab> usernameField    <tab><tab>       = DEFAULT_USERNAME_FIELD;

purplefox: @jviet can you help me with a translation here? :)

remi128: ok, i think i got it. Is it a question of formating the code? ( In the example above i can't see a difference )

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/UserFactory.java
purplefox: Sorry, I'm not sure I have understood the reasoning behind this... could you elaborate?

purplefox: Not sure what you mean by "But the interface of the User from vertx is defining only a minimum of the information and structure,"

remi128: Could be something like that:
https://github.com/remi128/apex-test/blob/master/src/main/java/User.java
This is not yet adapted onto the User-Interface from vertx, but will. In detail i am thinking about to build an O/R-Wrapper on top of mongo, first.
But if you think the factory is too much its no problem for me to remove it and to simply change that to the default user implementation.

purplefox: I'd prefer it to be removed.

vertx-auth-mongo/src/main/resources/io.vertx-vertx-mongo-service.json
```diff
@@ -0,0 +1,11 @@
+{
```
purplefox: Why is this needed?

vertx-auth-mongo/src/test/java/io/vertx/ext/auth/mongo/test/MongoBaseTest.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+  }
+
+  /**
+   * If instance of MongoService is null, initialization is performed
+   * 
+   * @return
+   * @throws Exception
+   */
+  public MongoService getMongoService() throws Exception {
```
purplefox: I don't understand why you are using MongoService - it would be much simpler to use MongoClient.

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
```diff
@@ -0,0 +1,299 @@
...
+
+  /**
+   * The default name of the property for the permissions, like it is stored in mongodb. Permissions are expected to be
+   * saved as JsonArray
+   */
+  String DEFAULT_PERMISSION_FIELD = "permissions";
+
+  /**
+   * The default name of the property for the username, like it is transported in credentials by method
+   * {@link #init(JsonObject)}
```
purplefox: error in javadoc

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
```diff
@@ -0,0 +1,299 @@
...
+   */
+  String DEFAULT_CREDENTIAL_USERNAME_FIELD = DEFAULT_USERNAME_FIELD;
+
+  /**
+   * The default name of the property for the password, like it is transported in credentials by method
+   * {@link #init(JsonObject)}
```
purplefox: error in javadoc

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
```diff
@@ -0,0 +1,299 @@
...
+   * The property name to be used to set the name of the field, where the salt style is stored inside
+   * 
+   * @see SaltStyle
+   */
+  String PROPERTY_SALT_STYLE = "saltStyle";
+
+  /**
+   * The property name to be used to set the name of the field, where the permissionsLookupEnabled is stored inside
+   */
+  String PROPERTY_PERMISSIONLOOKUP_ENABLED = "permissionsLookupEnabled";
```
purplefox: not used anywhere

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
```diff
@@ -0,0 +1,299 @@
...
+   * {@link #init(JsonObject)}
+   */
+  String DEFAULT_CREDENTIAL_PASSWORD_FIELD = DEFAULT_PASSWORD_FIELD;
+
+  /**
+   * The default name of the property for the salt field
+   */
+  String DEFAULT_SALT_FIELD = "salt";
+
+  String ROLE_PREFIX = "role:";
```
purplefox, line 134: missing javadoc

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
```diff
@@ -0,0 +1,299 @@
...
+
+  /**
+   * Creates an instance of MongoAuth
+   * 
+   * @param vertx
```
purplefox: missing javadoc for params

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
```diff
@@ -0,0 +1,299 @@
...
+  /**
+   * Create a {@link User} with the given parameters
+   * 
+   * @param username
+   * @param password
+   * @param roles
+   * @param permissions
+   * @return
+   */
+  public User createUser(String username, String password, List<String> roles, List<String> permissions);
```
purplefox: What's the purpose of this method? Users objects should only be created by the authenticate method.

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
```diff
@@ -0,0 +1,299 @@
...
+
+  /**
+   * Create a {@link User} with the given JsonObject
+   * 
+   * @param principal
+   * @return
+   */
+  public User createUser(JsonObject principal);
```
purplefox: As above

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
```diff
@@ -0,0 +1,449 @@
...
+import io.vertx.ext.mongo.MongoClient;
+
+import java.util.List;
+
+/**
+ * An implementation of {@link MongoAuth}
+ * 
+ * @author mremme
+ */
+public class MongoAuthImpl implements MongoAuth {
```
purplefox, line 40: Various errors in javadoc for this class. BTW you don't need to provide javadoc for internal classes, only public API

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuth.java
remi128: MongoAuth is using the data, which are stored inside a collection. That means, before somebody can authenticate, he has to be stored inside the mongo. The storing has to be done in a convenient way, following the properties of the MongoAuth. Means, the correct fields must be used and over all the password has to be crypted in the correct way, so that the authentication mechanism can deal with.
One solution is now, to document that and to explain somebody how a User must be stored in the MongoDb, so that the Auth is finding that. The other possibility is to offer a method  which simply does that. I prefer the second, cause the first one is much too unsecure in my mind and will take care for bugs.

purplefox: If you want to have a method that inserts some data into the db that's fine, but it can't return a User object as a User object must always correspond to an authenticated user.

remi128: ok, will modify that

remi128:
Hi Tim,

i modified now first the method createUser to insertUser with no return value. Please take a look to that and if this is ok i will fix the rest ( javadoc etc )

Mike

pmlopes:
Hi @remi128 this PR is not forgotten :) I will check it next week!

remi128:
good :-)

Issue #19, opened 2015-06-09
Suppress error stack of JWT trying to load unsupported MAC


panjiesw:
Hi,

Currently, as `JWT` class is trying to load MAC algorithm of configured keystore file, it logs NullPointerException if an algorithm isn't supported, as shown in [this](https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWT.java#L55) line. This makes the application logs to be polluted with rather meaningless error stack trace like so

```
22:49:23.968 [vert.x-eventloop-thread-2] io.vertx.ext.auth.jwt.impl.JWT [WARN ] [API] - HS256 not supported
java.lang.NullPointerException: null
    at io.vertx.ext.auth.jwt.impl.JWT.getMac(JWT.java:86) ~[vertx-auth-jwt-3.0.0-milestone6.jar:na]
    at io.vertx.ext.auth.jwt.impl.JWT.<init>(JWT.java:53) ~[vertx-auth-jwt-3.0.0-milestone6.jar:na]
    at io.vertx.ext.auth.jwt.impl.JWTAuthProviderImpl.<init>(JWTAuthProviderImpl.java:77) [vertx-auth-jwt-3.0.0-milestone6.jar:na]
    at io.vertx.ext.auth.jwt.JWTAuth.create(JWTAuth.java:39) [vertx-auth-jwt-3.0.0-milestone6.jar:na]
    at com.panjiesw.std.api.ApiVerticle.createAuthProvider(ApiVerticle.java:115) [main/:na]
    at com.panjiesw.std.api.ApiVerticle.startModule(ApiVerticle.java:88) [main/:na]
    at com.panjiesw.std.api.ApiVerticle.lambda$handleService$1(ApiVerticle.java:78) [main/:na]
    at com.panjiesw.std.api.ApiVerticle$$Lambda$8/383266732.handle(Unknown Source) [main/:na]
    at io.vertx.core.impl.DeploymentManager.lambda$reportResult$159(DeploymentManager.java:436) [vertx-core-3.0.0-milestone6.jar:na]
    at io.vertx.core.impl.DeploymentManager$$Lambda$14/677334899.handle(Unknown Source) [vertx-core-3.0.0-milestone6.jar:na]
    at io.vertx.core.impl.ContextImpl.lambda$wrapTask$15(ContextImpl.java:312) [vertx-core-3.0.0-milestone6.jar:na]
[...]

PR #20, opened 2015-06-10
Update JWT.java

pmlopes:
#19 do not throw NPE internally when the keystore does not contain all optional algorithms, instead properly assert and log as level INFO the missing optional algorithms.


pmlopes:
@purplefox, @vietj, @cescoffier please review

cescoffier:
Fine for me. Do you want me to merge it ?

Issue #23, opened 2015-06-15
Stormpath and vertx realm name disagree


gemmell:
This is the next issue related to #21, see that ticket for the setup.

```
protected String getAccountHref(PrincipalCollection principals) {
       Collection c = principals.fromRealm(getName());
       //Based on the createPrincipals implementation above, the first one is the Account href:
       return (String) c.iterator().next();
 }
```

The `getName()` returns `com.stormpath.shiro.realm.ApplicationRealm_0` and then in `principals.fromRealm` it does:

```
public Collection fromRealm(String realmName) {
    if (realmPrincipals == null || realmPrincipals.isEmpty()) {
        return Collections.EMPTY_SET;
    }
    Set principals = realmPrincipals.get(realmName);
    if (principals == null || principals.isEmpty()) {
        principals = Collections.EMPTY_SET;
    }
    return Collections.unmodifiableSet(principals);
}
```

Where the `realmPrincipals` is a map with one entry: `vertx-auth-shiro`

So it seems stormpath is looking up by principal using it's name, but somehow only `vertx-auth-shiro` has been registered.

New key for playing around if you didn't sign up yourself:

```
apiKey.id = 1Z1YUWJA1YVHAICAOIU8P96VW
apiKey.secret = Bb9pCd07rzCOlQ7uLr1ZZF2Q/zKQi8dfGHerMJ8fBE8
```

gemmell:
You can try to login with testname@test.com and pw 890fup*() and then do 

```
router.route().handler((RoutingContext context) -> {
    context.user().isAuthorised("role:0174870c-a06f-4337-900a-919b885f8a38+admins", (AsyncResult<Boolean> res) -> {
       if (res.succeeded()) {
          boolean hasPermission = res.result();
          if (hasPermission) {
             context.next();
          } else {
             context.response().setStatusCode(403).end("You do not have access to that group!");
          }
       } else {
          context.response().setStatusCode(403).end("Error asking if authorised");
       }
    });
 });
```

Note the group id is different to #21

pmlopes:
I am trying to reproduce the issue with a unit test but i cannot login, could you provide some working credentials or check the commit above

gemmell:
Hi Paulo if you want to keep it as a test case I suggest you create your own account at stormpath. I will be removing that key since it's my account which I may use for other things. It's really easy and free to sign up to stormpath.

You sign up
You create a user and give them a group
You create an API key
Done.

[...]

Issue #21, opened 2015-06-15
fromRealm returns null


gemmell:
As per my exploration of using vertx with shiro and stormpath here https://groups.google.com/forum/?fromgroups=#!topic/vertx/rO8xlDof1qA

I have encountered a defect. 

If in my code I do this:

```
  router.route().handler((RoutingContext context) -> {
        context.user().isAuthorised("role:84c2a410-b761-44cd-88c7-a8bba0e976cb+admins", (AsyncResult<Boolean> res) -> {
           if (res.succeeded()) {
              boolean hasPermission = res.result();
              if (hasPermission) {
                 context.next();
              } else {
                 context.response().setStatusCode(403).end("You do not have access to that group!");
              }
           } else {
              context.response().setStatusCode(403).end("Error asking if authorised");
           }
        });
     });
```

The `isAuthorised` calls through to stormpath's implementation of `doGetAuthorizationInfo`

```
@Override
protected AuthorizationInfo doGetAuthorizationInfo(PrincipalCollection principals) {
   assertState();
   SimpleAuthorizationInfo info = new SimpleAuthorizationInfo();
   String href = getAccountHref(principals);
...
protected String getAccountHref(PrincipalCollection principals) {
   Collection c = principals.fromRealm(getName());
   //Based on the createPrincipals implementation above, the first one is the Account href:
   return (String) c.iterator().next();
}
```

Which calls vert'x:

```
public class SimplePrincipalCollection implements PrincipalCollection {
...
@Override
public Collection fromRealm(String realmName) {
  return null;
}
```

And is hardcoded to null. And thus exceptions in getAccountHref

gemmell:
Hrrm. Looks to me like it's expecting a shiro `SimplePrincipalCollection`. The stormpath `ApplicationRealm.java` has the following function:

```
protected PrincipalCollection createPrincipals(Account account) {

    LinkedHashMap<String, String> props = new LinkedHashMap<String, String>();

    props.put("href", account.getHref());
    nullSafePut(props, "username", account.getUsername());
    nullSafePut(props, "email", account.getEmail());
    nullSafePut(props, "givenName", account.getGivenName());
    nullSafePut(props, "middleName", account.getMiddleName());
    nullSafePut(props, "surname", account.getSurname());

    Collection<Object> principals = new ArrayList<Object>(2);
    principals.add(account.getHref());
    principals.add(props);

    return new SimplePrincipalCollection(principals, getName());
}
```

[...]

PR #25, opened 2015-06-16
Feature/issue 23

pmlopes:
The issues #23 and #21 showed that the implementation of ShiroUser was not correct, this patch will make sure that the creation of shiro users is consistent across shiro plugins and that principals can be serialized over the wire.

There are 2 new unit tests showing a working integration with Stormpath and even using subjects serialized over the wire.

@purplefox @vietj @cescoffier please review


cescoffier:
Ok for me.

PR #27, opened 2015-08-03
Use JksOptions instead of raw JsonObject

pmlopes:
As discussed before, replace raw JsonObject config with JksOptions

### DIFF
--- a/vertx-auth-jwt/src/main/generated/io/vertx/rxjava/ext/auth/jwt/JWTAuth.java
+++ b/vertx-auth-jwt/src/main/generated/io/vertx/rxjava/ext/auth/jwt/JWTAuth.java
@@ -19,6 +19,7 @@ package io.vertx.rxjava.ext.auth.jwt;
 import java.util.Map;
 import io.vertx.lang.rxjava.InternalHelper;
 import rx.Observable;
+import io.vertx.core.net.JksOptions;
 import io.vertx.rxjava.core.Vertx;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.auth.jwt.JWTOptions;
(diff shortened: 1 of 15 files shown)


pmlopes:
@vietj please review

vertx-auth-jwt/src/main/generated/io/vertx/rxjava/ext/auth/jwt/JWTAuth.java
```diff
@@ -50,12 +51,32 @@ public Object getDelegate() {
    * @param config the config
    * @return the auth provider
    */
-  public static JWTAuth create(Vertx vertx, JsonObject config) { 
+  public static JWTAuth create(Vertx vertx, JksOptions config) {
```
cescoffier, line 54: This is breaking the API. The method should be kept and marked as deprecated.

vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/JWTAuth.java
```diff
@@ -37,11 +39,30 @@
    * @param config  the config
    * @return the auth provider
    */
-  static JWTAuth create(Vertx vertx, JsonObject config) {
+  static JWTAuth create(Vertx vertx, JksOptions config) {
```
cescoffier, line 42: This is breaking the API. The method should be kept and marked as deprecated.

I'm actually not sure we can just do a new JksOptions(config) ?

vertx-auth-jwt/src/test/java/io/vertx/ext/auth/test/jwt/JWTAuthProviderTest.java
```diff
@@ -38,11 +39,10 @@ public void setUp() throws Exception {
     authProvider = JWTAuth.create(vertx, getConfig());
   }
 
-  protected JsonObject getConfig() {
```
cescoffier, line 41: A test with the two way to create the authenticator would be nice (using JKSOptions and JsonObject)

pmlopes:
Since this would break the current API 3.x it will be postponed

Issue #26, opened 2015-07-30
Creating a new JWT token and adding user to session results in exceptions deserializing the JWT user object later


apatrida:
When clustered using Hazelcast, create JWT Token, validate it by calling provider.authenticate() then routingContext.setUser() so it is available immediately.  Next requests that come in all fail with:

```
19:19:56.440 [vert.x-eventloop-thread-2] ERROR i.v.e.w.impl.RoutingContextImplBase - Unexpected exception in route
com.hazelcast.nio.serialization.HazelcastSerializationException: Problem while reading DataSerializable, namespace: 0, id: 0, class: 'io.vertx.spi.cluster.hazelcast.impl.HazelcastAsyncMap$DataSerializableHolder', exception: Failed to load class io.vertx.core.VertxException: java.lang.InstantiationException: io.vertx.ext.auth.jwt.impl.JWTUser
    at com.hazelcast.nio.serialization.DataSerializer.read(DataSerializer.java:120) ~[hazelcast-3.5.jar:3.5]
    at com.hazelcast.nio.serialization.DataSerializer.read(DataSerializer.java:39) ~[hazelcast-3.5.jar:3.5]
    at com.hazelcast.nio.serialization.StreamSerializerAdapter.read(StreamSerializerAdapter.java:41) ~[hazelcast-3.5.jar:3.5]
    at com.hazelcast.nio.serialization.SerializationServiceImpl.toObject(SerializationServiceImpl.java:276) ~[hazelcast-3.5.jar:3.5]
    at com.hazelcast.spi.impl.NodeEngineImpl.toObject(NodeEngineImpl.java:200) ~[hazelcast-3.5.jar:3.5]
    at com.hazelcast.map.impl.MapServiceContextImpl.toObject(MapServiceContextImpl.java:281) ~[hazelcast-3.5.jar:3.5]
    at com.hazelcast.map.impl.proxy.MapProxySupport.toObject(MapProxySupport.java:1114) ~[hazelcast-3.5.jar:3.5]
    at com.hazelcast.map.impl.proxy.MapProxyImpl.get(MapProxyImpl.java:82) ~[hazelcast-3.5.jar:3.5]
    at io.vertx.spi.cluster.hazelcast.impl.HazelcastAsyncMap.lambda$get$13(HazelcastAsyncMap.java:46) ~[vertx-hazelcast-3.0.0.jar:na]
[...]

PR #28, opened 2015-08-03
#26 implement default constructor and warn in case of serializing the…

pmlopes:
… object, which is most likely a bad practice


cescoffier:
Ok for me.

PR #43, opened 2015-11-18
Issue 42

### LINKED ISSUE #42: Add an option to specify the signing algorithm
Actually the signing algorithm is the same as the certificate, but in some cases you must use another.

``` java
  CryptoSignature(final X509Certificate certificate, final PrivateKey privateKey) {
    this.certificate = certificate;
    this.privateKey = privateKey;
    try {
      this.sig = Signature.getInstance(certificate.getSigAlgName());
    } catch (NoSuchAlgorithmException e) {
      throw new RuntimeException(e);
    }
  }
```

In the "Google Cloud Plataform" we need to sign the JWT with **SHA256withRSA** using the private key of PKCS#12 certificate. The certificate is generated by Google and the sign algorithm of certificate is **SHA1withRSA** :(

https://developers.google.com/identity/protocols/OAuth2ServiceAccount

pmlopes:
Fixes #42: GCE certficates are signed with SHA1 while they expect they Tokens to be signed with SHA256


pmlopes:
It has been verified by reporter

PR #44, opened 2015-11-21
Fix example LDAP URL

jmoscato:
Please remove the extra colon from the example LDAP URL. Putting an extra colon as shown will cause the following exceptions to be silently thrown.

org.apache.shiro.authc.AuthenticationException: LDAP naming error while attempting to authenticate user.
...
Caused by: javax.naming.InvalidNameException: Invalid name: ://myldapserver.mycompany.com:10389


pmlopes:
Thanks for pointing that out

PR #46, opened 2015-11-27
update docs to state we need a unique index for mongo usernames column

### LINKED ISSUE #45: mongoAuth.insertUser creates multiple users with he same name
mongoAuth.insertUser creates multiple users with the same name causing authentication to fail.
i would expect that insertuser will validate that the same user does not exists.

pmlopes:
Fixes #45 

Document that mongo should have a index on username to avoid username duplicates


pmlopes:
This is a documentation only addition

PR #47, opened 2015-11-30
Provide options for jdbc and mongo and allow to create AuthProvider f…

vietj:
…rom the options with a lookup.

the goal is to create a convenient options from JSon, used for instance in shell service:

```
    vertx.deployVerticle("service:io.vertx.ext.shell", new DeploymentOptions().setConfig(new JsonObject().put("sshOptions",
        new JsonObject().
            put("host", "localhost").
            put("port", 5000).
            put("keyPairOptions", new JsonObject().
                put("path", "src/test/resources/server-keystore.jks").
                put("password", "wibble")).
            put("authOptions", new JsonObject().put("provider", "mongo").put("config",
                new JsonObject().
                    put("connection_string", "mongodb://localhost:27018")))))));
```

please @pmlopes @cescoffier review this PR.


vertx-auth-common/src/main/java/io/vertx/ext/auth/AuthOptions.java
```diff
@@ -16,26 +16,66 @@
...
+      case "mongo":
+        impl = "io.vertx.ext.auth.mongo.MongoAuthOptions";
+        break;
+      default:
+        throw new IllegalArgumentException("Invalid auth provider: " + provider);
+    }
 
-  public AuthOptions(JsonObject json) {
+    try {
+      ClassLoader cl = Thread.currentThread().getContextClassLoader();
```
cescoffier: Maybe null.

vertx-auth-common/src/main/java/io/vertx/ext/auth/AuthOptions.java
```diff
@@ -16,26 +16,66 @@
...
+      Class<?> optionsClass = cl.loadClass(impl);
```
cescoffier: The TCCL may be null, and may not be the right one (I hope it won't be able to load the class in this case). If the class cannot be loaded we should try with the current class classloader (that would delegate to the classpath classloader).

vertx-auth-common/src/main/java/io/vertx/ext/auth/AuthOptions.java
```diff
@@ -16,26 +16,66 @@
...
+   * @param json the json value
+   * @return the auth provider
+   */
+  static AuthOptions create(JsonObject json) {
 
-  public AuthOptions(AuthOptions that) {
-  }
+    String provider = json.getString("provider", "");
+    String impl;
+    switch (provider) {
```
cescoffier: No one can add its own provider ?

vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/JDBCAuthOptions.java
```diff
@@ -0,0 +1,177 @@
+package io.vertx.ext.auth.jdbc;
```
cescoffier, line 16: License header.

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/MongoAuthOptions.java
```diff
@@ -0,0 +1,247 @@
+package io.vertx.ext.auth.mongo;
```
cescoffier, line 16: License header.

vertx-auth-shiro/src/test/java/io/vertx/ext/auth/test/shiro/AuthOptionsTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ */
+public class AuthOptionsTest {
+
+  @Test
+  public void testSomething() {
```
cescoffier, line 17: testShiroOptionsInstantiation or whatever.

vertx-auth-common/src/main/java/io/vertx/ext/auth/AuthOptions.java
cescoffier: This would not work in an OSGi environment, as the bundle containing this class would not be able to load these classes.

(actually, I'm pretty sure it would not work with modules either)

vietj: well for now it would work would'n it ? or is there an obvious solution to this ?

vertx-auth-common/src/main/java/io/vertx/ext/auth/AuthOptions.java
vietj: at the moment it is not important, we can revisit that later.

vertx-auth-shiro/src/test/java/io/vertx/ext/auth/test/shiro/AuthOptionsTest.java
vietj, line 17: not possible, the class is not available.

vietj, line 17: but possible in dependant modules.

vertx-auth-common/src/main/java/io/vertx/ext/auth/AuthOptions.java
cescoffier: There is no obvious solution, the pattern need to be reversed. So, if we go this way, this will not pass modularization. So eventually we will need to rework this part in an incompatible way.

Unfortunately, there is no easy way to get it working as an SPI (with a hint-based lookup) may be a bit too heavy weight.

vertx-auth-common/src/main/java/io/vertx/ext/auth/AuthOptions.java
```diff
@@ -16,26 +16,66 @@
...
+    }
 
-  public AuthOptions(JsonObject json) {
+    try {
+      ClassLoader cl = Thread.currentThread().getContextClassLoader();
+      Class<?> optionsClass = cl.loadClass(impl);
+      Constructor<?> ctor = optionsClass.getConstructor(JsonObject.class);
+      return (AuthOptions) ctor.newInstance(json);
+    } catch (ClassNotFoundException e) {
+      throw new VertxException("Provider class not found " + impl + " / check your classpath");
```
cescoffier: it's not necessary in your classpath (especially with the TCCL).

PR #50, opened 2016-02-15
fix vert-x3/issues#100 source claims should be immutable

pmlopes:
see: https://github.com/vert-x3/issues/issues/100

### DIFF
--- a/vertx-auth-jwt/src/test/java/io/vertx/ext/auth/test/jwt/JWTAuthProviderTest.java
+++ b/vertx-auth-jwt/src/test/java/io/vertx/ext/auth/test/jwt/JWTAuthProviderTest.java
@@ -117,6 +117,18 @@ public class JWTAuthProviderTest extends VertxTestBase {
   }
 
   @Test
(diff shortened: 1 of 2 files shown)


vertx-auth-jwt/src/test/java/io/vertx/ext/auth/test/jwt/JWTAuthProviderTest.java
```diff
@@ -117,6 +117,18 @@ public void testGenerateNewToken() {
...
   @Test
+  public void testGenerateNewTokenImmutableClaims() {
+
+    JsonObject payload = new JsonObject()
+        .put("sub", "Paulo");
+
+    String token0 = authProvider.generateToken(payload, new JWTOptions().addPermission("user"));
+    String token1 = authProvider.generateToken(payload, new JWTOptions().addPermission("admin"));
+
+    assertNotSame(token0, token1);
```
cescoffier: Should it use assertNotEqual ?

PR #51, opened 2016-02-19
Replace html hyperlinks with asciidoc hyperlinks

slayful:


slayful:
The links in HTML are escaped in asciidoc.
http://vertx.io/docs/vertx-auth-oauth2/java/#_the_oauth2_auth_provider

cescoffier:
The asciidoc file is generated from the package-info.java file. Could you fix the issue in this file instead ?

pmlopes:
Fixed at the source

PR #55, opened 2016-02-28
fixes #54

### LINKED ISSUE #54: JWT Signature verification failed with multithread
sometimes JWT.decode will throw the RuntimeException("Signature verification failed") when we use blockingHandler or generateToken/authenticate in different thread.

I think it's because CryptoMac use mac.doFinal to generate digests and it's not [thread-safe.](http://stackoverflow.com/questions/31898802/is-macdofinal-thread-safe-in-java)

davidafsilva:
A simple intrinsic lock approach should fix the problem.


pmlopes:
Thanks!

PR #57, opened 2016-03-02
fix #56: allow full control of the oauth2 httpclient by extending Htt…

### LINKED ISSUE #56: oauth2 should allow all config from httpclient
- use case allow trust all ssl certificates for internal server

pmlopes:
…pClientOptions

fix #56

This will create a breaking change since we cannot have both a constructor with JsonObject and Options object (codegen will fail for JS)


vietj:
:+1:

PR #58, opened 2016-03-03
Fix #53: get keycloak openid-connect (oauth2) better integrated with …

### LINKED ISSUE #53: implement keycloak provider
see: https://github.com/vert-x3/wiki/wiki/KeyCloak-integration

pmlopes:
…auth.

Fix #53


pmlopes:
This is the auth part of the integration. (a seperate PR needs to be done to web).

In order to reuse code what this PR does is add a DataObject that converts `keycloak.json` config file to the generic OAuth2 options and from there it behaves just like any other oauth2 provider.

Since keycloak is not really oauth2 but openid-connect there is extra logic to parse the tokens which are known to be in JWT format, this code is not keycloak specific and can be enabled for other providers using the option `setJwtToken(true)`.

Finally since openid-connect has the concept of logout the AccessToken object is now extended to also handle logging out (ending sessions).

vertx-auth-oauth2/src/main/asciidoc/groovy/index.adoc
```diff
@@ -382,21 +382,18 @@ def oauth2 = OAuth2Auth.create(vertx, OAuth2FlowType.CLIENT, credentials)
...
+----
 
+When using this approach the provider has knowledge on how to parse access tokens and extract grants from inside.
+This information is quite valuable since it allows to do authorization at the API level, for example:
 
-// Initialize the OAuth2 Library
-def oauth2 = OAuth2Auth.create(vertx, OAuth2FlowType.CLIENT, credentials)
-
+[source,groovy]
+----
```
cescoffier, line 415: Codetrans issue. The code you have added cannot be translated.

vertx-auth-oauth2/src/main/asciidoc/js/index.adoc
```diff
@@ -373,20 +373,18 @@ var oauth2 = OAuth2Auth.create(vertx, 'CLIENT', credentials);
...
-// Set the client credentials and the OAuth2 server
-var credentials = {
-  "clientID" : "CLIENT_ID",
-  "clientSecret" : "CLIENT_SECRET",
-  "site" : "https://www.your-keycloak-server.com",
-  "authorizationPath" : "/realms/" + realm + "/protocol/openid-connect/auth",
-  "tokenPath" : "/realms/" + realm + "/protocol/openid-connect/token"
-};
-
+Code not translatable
```
cescoffier: Codetrans issue.

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
```diff
@@ -38,17 +37,27 @@
 
   private static final Logger log = LoggerFactory.getLogger(AccessTokenImpl.class);
 
-  private Vertx vertx;
-  private OAuth2ClientOptions config;
+  private static final JsonObject EMPTY_JSON = new JsonObject();
```
cescoffier, line 40: (not related to this issue, but more general)

I wonder if we should defines these object in the (core) Json class. However, I don't see a way to be sure no-one modify them (so they are not empty anymore). Maybe not a good idea...

cescoffier:
A couple of codetrans issue. You should have a log file with some details on the error.

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
pmlopes, line 40: yes we could add some constants like:

```
EMPTY = new JsonObject(Collections.unmodiffiableMap(Collections.emptyMap()));
```

vertx-auth-oauth2/src/main/asciidoc/js/index.adoc
pmlopes: The issue there is that I really want to use the json string since it is a copy paste from the keycloak config webpage. I can translate it to code but some people might think that they need to code it instead of just import from the admin page.

PR #60, opened 2016-03-18
Fixes #59: allow read only token generation based on public keys

### LINKED ISSUE #59: Feature: Do not require private key for JWT if not planning to sign tokens?
Right now the JWT module requires a private key for asymetric keys in order to be used, however, if you're only interested in decoding and verifying a key then this is an unnecessary additional piece of security data that needs to be stored in the key store. The piece of code responsible for this is located here:

``` java
X509Certificate certificate = getCertificate(keyStore, alg);
PrivateKey privateKey = getPrivateKey(keyStore, keyStorePassword, alg);
if (certificate != null && privateKey != null) {
  tmp.put(alg, new CryptoSignature(alias.get(alg), certificate, privateKey));
} else {
  log.info(alg + " not available");
}
```

Can the requirement that a private key be present be lifted?

pmlopes:
Fixes #59 

generate tokens using a public key. These tokens cannot be encoded on the host but on the node that has the private key.

### DIFF
--- a/vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWT.java
+++ b/vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWT.java
@@ -24,6 +24,8 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.security.*;
 import java.security.cert.X509Certificate;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.X509EncodedKeySpec;
 import java.util.*;
 
 /**
@@ -39,6 +41,10 @@ public final class JWT {
 
   private final Map<String, Crypto> CRYPTO_MAP;
   private final boolean unsecure;
+
+  public JWT() {
+    this(null);
+  }
 
   public JWT(final KeyStore keyStore, final char[] keyStorePassword) {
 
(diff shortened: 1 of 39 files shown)


pmlopes:
Still need to add documentation

vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWT.java
```diff
@@ -40,6 +42,10 @@
   private final Map<String, Crypto> CRYPTO_MAP;
```
vietj: why upper case ?

vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWTAuthProviderImpl.java
```diff
@@ -64,7 +64,15 @@ public JWTAuthProviderImpl(Vertx vertx, JsonObject config) {
 
         this.jwt = new JWT(ks, keyStore.getString("password").toCharArray());
       } else {
-        this.jwt = new JWT(null, null);
+        // in the case of not having a key store we will try to load a public key in pem format
+        // this is how keycloak works as an example.
+        String publicKey = config.getString("public-key");
+
+        if (publicKey != null) {
```
vietj: given that JWT() calls JWT(null), the test does not seem to make much a difference

pmlopes:
Documentation added and example too

PR #63, opened 2016-04-18
Handle plain/text responses as url encoded

### LINKED ISSUE #62: OAuth2API can't handle plain text responses
### Version
- vert.x core: 3.2.1
- vert.x auth: 3.2.1
### Context

I was trying to get an access token from facebook and noticed it returns text/plain which OAuth2API cannot handle.  The error thrown is:

Access Token Error: Cannot handle content type: text/plain
### Steps to reproduce

Try to use the configuration described here to get an access token from facebook:
http://vertx.io/docs/vertx-auth-oauth2/java/
### Extra

Here's someone else that ran into this:
http://stackoverflow.com/questions/35245955/vert-x-oauth2-facebook-api-call-to-get-access-token-fails-because-of-text-plain

pmlopes:
Fixes #62 

Because Facebook does not return the right content type


vietj:
👍

PR #78, opened 2016-08-10
JWT audience can be a array or just single string value, and we shoul…

manish-panwar:
If client is sending the single "aud" then Vertx-JWT doesn't handle it gracefully.

It's not always necessary for the "aud" to be a JSON array, and it can be just a single string value.
Nowhere in JWT specs I found that "aud" always has to be an array of String.

See https://groups.google.com/forum/#!topic/vertx/7hFYACYMLQs 

Signed-off-by: manishk


manish-panwar:
@vietj  Is this something we can merge to our upcoming release ? We need this and we don't want to build a custom solution in our code and rather use vertx-jwt API.

PR #79, opened 2016-08-11
Add touch method to the User interface to prevent ShiroUsers from timing out every 30 min

johnoliver:
Currently the `lastAccessTime` on the `subject.session` (an instance of `SimpleSession`) that is inside the `ShiroUser` object is never updated. This means that all Users will timeout after 30 min even if they are active. There is currently no way to reset the timeout for these users.

This adds `touch` to the `User` interface that allows the user to indicate they are still active. It does unfortunately add a method that is reasonably Shiro specific to the User interface, hence why the other Users touch methods are NOOP. But this seemed about the cleanest way to allow users to indicate they are still active.

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
@@ -280,6 +280,11 @@ public class AccessTokenImpl extends AbstractUser implements AccessToken {
   }
 
   @Override
+  public void touch() {
+    //NOOP
+  }
+
+  @Override
   public void writeToBuffer(Buffer buff) {
     super.writeToBuffer(buff);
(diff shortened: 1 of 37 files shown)


vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
```diff
@@ -280,6 +280,11 @@ public void setAuthProvider(AuthProvider authProvider) {
   }
 
   @Override
+  public void touch() {
+    //NOOP
```
pmlopes, line 284: This is tricky... in oauth2 a touch can also be seen as a call to a refresh token

pmlopes:
I'm a bit afraid of this change since it will propagate to the other mechanisms that have their own touch semantics like oauth2 and others like jwt do not support touch since the token itself is issues with an optional expiration date and it requires a new token if refreshed.

pmlopes:
Shiro is deprecated on 4.0.0 so we're closing this issue.

PR #84, opened 2016-08-30
Add helper class for Keycloak

sczyh30:
Add a helper class for Keycloak so that we can easily retrieve decoded token and some necessary data (e.g. `preferred_username`) from the Keycloak principal.

Signed-off-by: sczyh30 sczyh16@gmail.com

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/KeycloakHelper.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/KeycloakHelper.java
@@ -22,7 +22,7 @@ public interface KeycloakHelper {
    * @param principal user principal
    * @return the raw id token string
    */
-  static String getRawIdToken(JsonObject principal) {
+  static String rawIdToken(JsonObject principal) {
     return principal.getString("id_token");
   }
 
@@ -32,8 +32,8 @@ public interface KeycloakHelper {
    * @param principal user principal
    * @return the id token
    */
-  static JsonObject getIdToken(JsonObject principal) {
-    return parseToken(getRawIdToken(principal));
+  static JsonObject idToken(JsonObject principal) {
+    return parseToken(rawIdToken(principal));
   }
 
   /**
@@ -42,7 +42,7 @@ public interface KeycloakHelper {
    * @param principal user principal
    * @return the raw access token string
    */
-  static String getRawAccessToken(JsonObject principal) {
+  static String rawAccessToken(JsonObject principal) {
     return principal.getString("access_token");
   }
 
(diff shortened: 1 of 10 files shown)


vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/KeycloakHelper.java
```diff
@@ -0,0 +1,143 @@
...
+      return null;
+    }
+    String[] parts = token.split("\\.");
+    if (parts.length < 2 || parts.length > 3) {
+      throw new IllegalArgumentException("Parsing error");
+    }
+    return new JsonObject(decodeBase64(parts[1])); // get "payload" part
+  }
+
+  private static String decodeBase64(String s) {
```
pmlopes: You shouldn't use the sun package, on java8 there is a Base64 encoder/decode under java.util. It has support for all encoding types so i guess you can even skip the padding/escape code by selecting the right one.

see: https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/KeycloakHelper.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Helper class for processing Keycloak principal.
+ *
+ * @author Eric Zhao
+ */
+public final class KeycloakHelper {
```
pmlopes: This should be a interface so we could have it generated in all languages through codegen

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/KeycloakHelper.java
```diff
@@ -0,0 +1,143 @@
+package io.vertx.ext.auth.oauth2;
```
pmlopes, line 17: Wouldn't this make more sense under the oauth2 code?

sczyh30: Do you mean the code should be put in `OAuth2Auth` interface or in other places?

pmlopes: I mean that it probably makes more sense under this project:
https://github.com/vert-x3/vertx-auth/tree/master/vertx-auth-oauth2

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/KeycloakHelper.java
```diff
@@ -0,0 +1,143 @@
...
+    return principal.getString("access_token");
+  }
+
+  /**
+   * Get decoded `access_token` from the principal.
+   *
+   * @param principal user principal
+   * @return the access token
+   */
+  public static JsonObject getAccessToken(JsonObject principal) {
```
pmlopes: for consistency with the rest of the API's if a propery has only a getter then we usually omit the `get` prefix, so it should be something like `accessToken(JsonObject principal)`

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/KeycloakHelper.java
sczyh30: Yes I've put it in the `vertx-auth-oauth2` , and modified names, use Java 8 Base64 lib instead as well as add polyglot language support.

PR #95, opened 2016-10-25
Use getEncoder instead of getUrlEncoder

mguillet:
According to RFC7617 I don't believe we need to encode special character before doing base64 encoding.

I had an issue with some credentials that contained '$' in them producing a different output with `Base64.getEncoder()`. My OAuth provider didn't accept the authorisation header with `Base64.getUrlEncoder()` but it worked with `Base64.getEncoder()`

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2API.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2API.java
@@ -67,7 +67,7 @@ public class OAuth2API {
       params.remove("access_token");
     } else if (config.isUseBasicAuthorizationHeader() && config.getClientID() != null && !params.containsKey("client_id")) {
       String basic = config.getClientID() + ":" + config.getClientSecret();
-      headers.put("Authorization", "Basic " + Base64.getUrlEncoder().encodeToString(basic.getBytes()));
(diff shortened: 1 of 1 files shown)


pmlopes:
See https://www.rfc-editor.org/rfc/rfc7515.txt sectio 3.1 it specifies Base64 URL encoding. Can you share which consumer are you trying to use to see what don't they support this?

mguillet:
Paulo, the RFC7515 is about JWS isn't it ?

Sorry if I wasn't clear, I did put very little background in my PR.

I encountered the issue when doing an OAuth Client Credentials Flow with my provider.

Let's say the `clientID` and `clientSecret` are something like:
- clientId: `myGreatApp$`
- clientSecret: `myGreatSecret`

When crafting the Basic Auth header it gives us:
- `Base64(clientId + ":" + clientSecret) => bXlHcmVhdEFwcCQ6bXlHcmVhdFNlY3JldA==`
- `Base64UrlEnc(clientId + ":" + clientSecret) => bXlHcmVhdEFwcCUyNCUzQW15R3JlYXRTZWNyZXQ=`

Vert.x is doing url + Base64 encoding and my provider is expecting just base64 encoding.

Let's me know if you need other details.

pmlopes:
Ok sorry I thought the issue was related to the Authorization bearer url and I assumed jwt for some strange reason :-/

mguillet:
Thanks!

Issue #88, opened 2016-10-03
Allow using Google JWT directly (by implementing server-to-server communication)


pmlopes:
Google allows applications to communicate securely using just JWTs over its OAuth2 interface. This is also known as server to server communication and avoids the OAuth2 redirect flow.

We could already do this with the current JWT module but we could add an helper to simplify its usage.

This mode of communication is very useful for consuming APIs from google for example consuming good drive files.

pmlopes:
Reference: https://developers.google.com/identity/protocols/OAuth2ServiceAccount
Mentor: Paulo Lopes

PR #106, opened 2016-11-23
Issues/better oauth2 experience

pmlopes:
This PR will make the developer experience easier when dealing with oauth

Users only need to know the configuration that their provider assigns to them

* client id
* client secret

The helpers will configure the default urls, scope separators, etc...

From the old list we can now off the box use:

* App.net
* Azure AD
* Box.com
* Dropbox
* Facebook
* Foursquare
* Github
* Google
* Google Server to Server
* Heroku
* Instagram
* Keycloak
* Linkedin
* Live.com
* Mailchimp
* Salesforce
* Shopify
* Soundcloud
* Stripe
* Twitter

The usage is now as simple as:

1. Choose a provider e.g.: google
2. Get the client id and client secret from your provider
3. use `Google.create(vertx, clientId, secretId)`

Before many configuration was required now that info has been pre-compiled for the end user

JWT tokens now share the same JWT library from JWTAuth instead of having it duplicated and out of sync. The JWT library has been expanded to support Private Key initializers required for RFC 7523 (server to server oauth2).


vietj:
why can't we have all the static methods in the same OAuth2Auth class but with methods like createInstagramAuth + it would allow the user to have them all appear when doing code completion in the IDE.

pmlopes:
We could but i though that having a single class for each provider would make things cleaner as for example if google changes their API one only needs to update 1 file and be sure that it won't affect others.

An evolutionary step here will be adding support to fetch the provider user object (currently we take the token and that is our user) but say that you're working with google you might want to use their API and once you do the same for Facebook then you need to use its graph API (they are all different).

Having the providers in their own factory would allow adding provider specific features like that without the need to write generic compatibility methods.

vietj:
ok

pmlopes:
This PR will also address and fix: vert-x3/vertx-web#467

pmlopes:
Also fixes #88

PR #110, opened 2016-12-16
implemented RFC7662 token introspection

### LINKED ISSUE #109: Add Oaut2 access_token validation to the existing providers.
@javadevmtl commented on [Tue Oct 18 2016](https://github.com/vert-x3/issues/issues/208)

Based on the conversation with Paulo: https://groups.google.com/forum/#!topic/vertx/tGgpL4bpdqo

Currently the OAuth2 providers only issue access_tokens, they do not validate access tokens. So according to Paulo an improvement can be made to make this possible.

Took a bit of google/hunting down, but to validate an access_token with the Keycloack provider you are required to make an API call to:

```
oauth2.api(HttpMethod.POST, "/realms/{YOUR_REALM_HERE}/protocol/openid-connect/token/introspect", new JsonObject().put("token", ACCESS_TOKEN_HERE), h -> {});
```

---

@pmlopes commented on [Wed Nov 23 2016](https://github.com/vert-x3/issues/issues/208#issuecomment-262522342)

We will need an extra method in the providers to validate token, of course this is different across providers so some work is needed here.

* google: https://www.googleapis.com/oauth2/v1/tokeninfo?access_token=1/fFBGRNJru1FQd44AzqT3Zg
* keykloak: /realms/{YOUR_REALM_HERE}/protocol/openid-connect/token/introspect?token=access_token
* etc...

---

@javadevmtl commented on [Thu Nov 24 2016](https://github.com/vert-x3/issues/issues/208#issuecomment-262848775)

Here is more info if you like: https://keycloak.gitbooks.io/authorization-services-guide/content/topics/service/protection/token-introspection.html

pmlopes:
Fixes #109

Implements a way to introspect tokens, both the official RFC and the non official Google Way

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
@@ -144,14 +144,17 @@ public class AccessTokenImpl extends AbstractUser implements AccessToken {
   @Override
   public boolean expired() {
 
+    long now = System.currentTimeMillis();
+
+    // expires_at is a computed field always in millis
+    if (token.containsKey("expires_at") && token.getLong("expires_at", 0L) < now) {
+      return true;
+    }
+
     // All dates in JWT are of type NumericDate
     // a NumericDate is: numeric value representing the number of seconds from 1970-01-01T00:00:00Z UTC until
     // the specified UTC date/time, ignoring leap seconds
-    final long now = System.currentTimeMillis() / 1000;
-
-    if (token.containsKey("expires_at") && token.getLong("expires_at", 0L) < System.currentTimeMillis()) {
-      return true;
-    }
+    now = System.currentTimeMillis() / 1000;
 
     if (content != null) {
       if (content.containsKey("exp")) {
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
@@ -29,6 +29,10 @@ import io.vertx.ext.auth.AuthProvider;
 import io.vertx.ext.auth.oauth2.AccessToken;
 
 import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.regex.Pattern;
+
+import static io.vertx.ext.auth.oauth2.impl.OAuth2API.api;
 
 /**
  * @author Paulo Lopes
@@ -85,10 +89,43 @@ public class AccessTokenImpl extends AbstractUser implements AccessToken {
     }
 
     this.token = json;
+    this.content = null;
+
+    // init from introspection (in this case we bypass the decode)
+    if (token.containsKey("introspect")) {
+      content = token.getJsonObject("introspect");
+      token.remove("introspect");
+
+      // the permission cache needs to be clear
+      clearCache();
+      // rebuild cache
+      if (content.containsKey("scope")) {
+        String[] scopes = content.getString("scope", "").split(Pattern.quote(provider.getScopeSeparator()));
+        Collections.addAll(cachedPermissions, scopes);
+        // not JWT property
+        content.remove("scope");
+      }
+      // validate expiration
+      if (content.containsKey("exp")) {
+        Long exp;
+        try {
+          exp = content.getLong("exp");
+        } catch (ClassCastException e) {
+          // for some reason someone decided to send a number as a String...
+          exp = Long.valueOf(content.getString("exp"));
+        }
+        token.put("expires_at", 1000 * exp);
+      }
+    }
 
     // try to parse the access_token
     if (provider.getConfig().isJwtToken() && json.containsKey("access_token")) {
-      content = provider.decode(json.getString("access_token"));
+      final JsonObject jwt = provider.decode(json.getString("access_token"));
+      if (content != null) {
+        content.mergeIn(jwt);
+      } else {
+        content = jwt;
+      }
     }
   }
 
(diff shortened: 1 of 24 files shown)


vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
```diff
@@ -84,20 +88,96 @@ private void init(JsonObject json) {
...
   @Override
   public boolean expired() {
-    return token.containsKey("expires_at") && token.getLong("expires_at", 0L) < System.currentTimeMillis();
+
+    // All dates in JWT are of type NumericDate
+    // a NumericDate is: numeric value representing the number of seconds from 1970-01-01T00:00:00Z UTC until
+    // the specified UTC date/time, ignoring leap seconds
+    final long now = System.currentTimeMillis() / 1000;
+
+    if (token.containsKey("expires_at") && token.getLong("expires_at", 0L) < System.currentTimeMillis()) {
```
tsegismont: So `expires_at` is milliseconds since epoch, but the rest (`exp`) is second since epoch? Or is the comparison to `System.currentTimeMillis` (instead of `now`) a bug?

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
```diff
@@ -84,20 +88,96 @@ private void init(JsonObject json) {
...
+      if (content.containsKey("exp")) {
+        if (now >= content.getLong("exp")) {
+          return true;
+        }
+      }
+
+      if (content.containsKey("iat")) {
+        Long iat = content.getLong("iat");
+        // issue at must be in the past
+        if (iat > now) {
```
tsegismont, line 169: Won't this be always true? It seems strange that a system would generate a "future" issue date.

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
```diff
@@ -189,6 +269,53 @@ public AccessTokenImpl logout(Handler<AsyncResult<Void>> callback) {
...
+            // reset the access token
+            init(new JsonObject().put("access_token", token).mergeIn(json));
+
+            if (expired()) {
+              callback.handle(Future.failedFuture("Expired token"));
+              return;
+            }
+
+            callback.handle(Future.succeededFuture());
+          } catch (RuntimeException e) {
```
tsegismont, line 308: What kind of errors are you trying to defend against with those two try/catch blocks?

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/flow/OAuth2Flow.java
```diff
@@ -37,4 +37,9 @@ default String authorizeURL(JsonObject params) {
    * See implementations for specific documentation
    */
   void getToken(JsonObject params, Handler<AsyncResult<AccessToken>> handler);
+
+  /**
+   * See implementations for specific documentation
```
tsegismont: Minor but Javadoc requires to have the first sentence ending with a dot.

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/package-info.java
```diff
@@ -187,6 +187,31 @@
...
+ *
+ * === Verifying JWT tokens
+ *
+ * We've just covered how to introspect a token however when dealing with JWT tokens one can reduce the amount of trips
+ * to the provider server thus enhancing your overall response times. In this case tokens will be verified using the
+ * JWT protocol at your application side only.
+ *
+ * [source,$lang]
+ * ----
+ * {@link examples.AuthOAuth2Examples#example16}
```
tsegismont, line 215: I'm not familiar with JWT. Does this mean that the request implied by the call to `decodeToken` will perform the introspection as well?

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/package-info.java
```diff
@@ -187,6 +187,31 @@
  * The provider also supports Server to Server or the RFC7523 extension. This is a feature present on Google with their
  * service account.
  *
+ * === Token Introspection
+ *
+ * Tokens can be introspected in order to assert that they are still valid, although there is RFC7660 for this purpose
```
tsegismont: I'd cut the sentence into two. "Tokens can be introspected in order to assert that they are still valid**. A**lthough"

tsegismont (changes requested):
This test fails on my box:

```
Tests run: 6, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.207 sec <<< FAILURE! - in io.vertx.ext.auth.test.oauth2.OAuth2AccessTokenTest
introspectAccessToken(io.vertx.ext.auth.test.oauth2.OAuth2AccessTokenTest)  Time elapsed: 0.028 sec  <<< ERROR!
java.lang.NullPointerException: null
	at io.vertx.ext.auth.oauth2.impl.OAuth2API.api(OAuth2API.java:43)
	at io.vertx.ext.auth.oauth2.impl.flow.CommonFlow.introspectToken(CommonFlow.java:60)
	at io.vertx.ext.auth.oauth2.impl.flow.AuthCodeImpl.introspectToken(AuthCodeImpl.java:33)
	at io.vertx.ext.auth.oauth2.impl.OAuth2AuthProviderImpl.introspectToken(OAuth2AuthProviderImpl.java:152)
	at io.vertx.ext.auth.oauth2.impl.OAuth2AuthProviderImpl.introspectToken(OAuth2AuthProviderImpl.java:124)
	at io.vertx.ext.auth.test.oauth2.OAuth2AccessTokenTest.introspectAccessToken(OAuth2AccessTokenTest.java:147)
```

Other than that, the changes look good to me.

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
pmlopes: This is correct, though probably I should reuse the now variable. "expires_at" is a computed variable not part of the spec, When I get a response from the server it may contain "expires_in" (in seconds) so at that moment I compute "expires_at" as current time millis + 1000 * expires_in and it is always millis.

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
pmlopes: clock synch'ing issues/replay attacks...

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
pmlopes: json parsing errors, for example, Azure does not follow the spec and returns expires_in as a String instead of a number, if someone sends say a boolean it would throw a JSON decode exception

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/package-info.java
pmlopes: no, JWT are stateless so they can be decoded on our server only, no need to ask the provider to verify if the token is valid. HOWEVER :) due to the nature of JWT there is no concept of logout so a user might be logged out but the token is still valid to be used since it hasn't expired. For those cases one needs to use the introspection API and ask the provider, is this token still valid?

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java
pmlopes: the iat in the future can also be used to avoid being logged out. For example you know your token will expire in 5 minutes you might already request a new one that the server will say that will only valid in 5 minutes, if there is a MITM attack the new token is not vulnerable until it is in use.

pmlopes:
Test has been fixed.

tsegismont (approved):

Issue #102, opened 2016-11-13
MongoAuth: allow storing additional fields when creating users


DieterDePaepe:
### Version

* vert.x core: 3.3.3
* vert.x auth: 3.3.3

### Context

MongoAuth provides a method to add a new User: 

    void insertUser(String username, String password, List<String> roles, List<String> permissions,
      Handler<AsyncResult<String>> resultHandler);

This approach does not provide any way to supply additional fields to the entry being saved. I suggest changing this method to add a JsonObject to which all data will be added before being stored:

    void insertUser(JsonObject template, String username, String password, List<String> roles, List<String> permissions,
      Handler<AsyncResult<String>> resultHandler);

stefanvitz:
+1

stefanvitz:
What is the the status here? 
When the team sees the value added on this one perhaps we could prepare an pull request.
the alternative is to work around in my code to insert a user via MongoAuth and then manually update the user with the needed data... (as any other user would have to do),,,

pmlopes:
I'd say that if you're passing a json object template the template itself should already contain the username and password:

```java
void insertUser(JsonObject template, List<String> roles, List<String> permissions,
  Handler<AsyncResult<String>> resultHandler);
```

WDYT?

pmlopes:
@stefanvitz feel free to discuss and make PRs

DieterDePaepe:
@pmlopes I'd disagree for 2 reasons.

1) username and password are 2 things that MongoAuth tracks as special fields (MongoAuth knows which key they are stored under). So why merge username and password with the template, but leave out permissions and roles? Also, this would require the caller to also know the keys under which username and password are stored.

2) the password shouldn't be stored in the JSON. MongoAuth takes care of the hashing strategy and only stores the hash. Sure, it could replace the password out of the template and replace it with the hash, but that seems more error prone.

I'd suggest something like

```
void insertUser(JsonObject template, String username, String password, List<String> roles, List<String> permissions,  Handler<AsyncResult<String>> resultHandler) {
  //Throw error if username or password field are already specified in template
  //Throw error if roles/permissions are not null and are already specified in template
  //Functionality as it exists now.
}
```

I'm willing to make a PR if you agree with this approach and if you let me know what to do with the existing method? (Adjust it, remove it, leave it be)

pmlopes:
[...]

PR #111, opened 2016-12-18
MongoAuth: Allow custom data when creating new user.

DieterDePaepe:
Issue #102.


pmlopes:
I believe we should pass the ‘IllegalArgumentException‘ to the handler instead of throwing to avoid try catch blocks while using the API.

stefanvitz:
I agree with pmlopes. the change looks fine but an handle(Future.failedFuture..)  would most likely be my choice here as well.

DieterDePaepe:
I'd disagree. This kind of error is a programming error and not a program execution error, and therefor an immediate IllegalArgumentException seems best. I'd imagine any programmer keeping some kind of schema of their data model (of what is being put into mongo), so it is at this stage that they should take care that no fields overlap. They shouldn't be caught at runtime, but discovered in advance using tests.

pmlopes:
You can still return the typed exception in the future:

```java
Future.failedFuture(new IllegalArgumentException(reason));
```

this makes the code cleaner in the sense you only need:

```
mongo.createUser( ... res -> {
  if (res.failed()) {
    // do something, probably inspect the type of the cause
    if (res.cause() instanceof IllegalArgumentException) {
      // handle this is a special way...
    }
  } else {
    // everything went fine!
  }
});
```

Instead of:

```java
try {
  mongo.insertUser(... res -> {
    if (res.failed()) {
      // handle error (1)
    } else {
      // everything went fine!
    }
  });
} catch (IllegalArgumentException e) {
  // handle error (2) for this case...
}
```

In the second case you need to handle errors in 2 places using 2 different notations and it can be tricky once you have a bunch of async calls and users get confused if the top level try catch will catch errors from the handlers (which it won't).

DieterDePaepe:
You may have misunderstood me. I'm arguing that this is a type of error that shouldn't be caught at all, similar as how you would not write checks for NullPointerException everywhere in the code. If this error occurs, it is a programming bug, not a recoverable error.

Further argumentation why it is a programming bug, based on my own usage of this library: I am storing users in mongo, the structure of a user in the mongo collection is my responsibility. Therefor I should make that there are no conflicting keys between my own structure of a user and the username/password/roles/permissions fields that I am passing to MongoAuth.

See also the responses to [this stackoverflow question](http://stackoverflow.com/questions/15208544/when-should-an-illegalargumentexception-be-thrown).

pmlopes:
I understand the reasoning of runtime exceptions and agree however due to the async model of vert.x if you throw a runtime exception it won't unwind the stack to a state your application crashes or your current request is terminated but leaves it in a intermediate state since no one will call the request to terminate. In these cases what will happen is that a request will stall until there is a socket timeout and you won't have any way to either terminate it or handle it.

By passing it to the future you can later on ignore it or handle it as you want...

stefanvitz:
ok... so your argumentation here would be: the programmer has to ensure the right format of the json before passing it. 
im not quite sure. i assume it depends how you create the template json object. If it is dynamicly created the programmer perhaps does not know the exact content and the exception even must not happen every time the code is executed... and perhaps not even  while you are codeing your application until way later.... 
Not catching the IllegalArgumentException could cause hassle in your code... resulting in an handled rest call for example to not complete at all,
while the usage of the handler would result in an failure that you could handle normally..

Normally i just want my defined flow to complete flawlessy...  an unchecked exception because the content of an json object was not checked can really mess that up. So in the end you would force the programmer to check the IAE so that the request ALWAYS completes.

pmlopes:
You can still write code assuming that there are no errors like:

```java
mongo.createUser( ... res -> {
  // everything went fine (maybe)! 
  res.result().someMethod(); // might throw NPE if there was a IllegalArgumentException during creation
});
```

DieterDePaepe:
@stefanvitz how does that differ from having to check whether you're not going to run into a NullPointerException for some variable that you dynamically determined? Just as the programmer might know that a certain method returns null (and having to check that before continuing to use that value), a programmer should know that a certain method might return keys that overlap with the predefined username/password/... keys.

@pmlopes for the last code fragment, you'd also get a timeout in case of an error, right? (And anyway, you cannot assume everything went right at that point, since problems might occur with the database. )

DieterDePaepe:
Any updates on this? I still prefer the original way I implemented this, but I'm willing to make the change to get this PR accepted. I think I've made my point, so I'll leave the final decision to you guys.

pmlopes:
@DieterDePaepe you should use the async result way instead of exception because it will help also how the codegen project will then work out the API to be usable from `RX`, `CompletableFuture`s and so on. By keeping it in that style users who prefer RX will get the new API for "free" thanks to codegen.

(sorry for a late comment, was on vacation until yesterday ;))

DieterDePaepe:
Updated as requested.

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
```diff
@@ -163,31 +163,31 @@ public void insertUser(String username, String password, List<String> roles, Lis
 
   @Override
   public void insertUser(JsonObject principal, String username, String password, List<String> roles,
-      List<String> permissions, Handler<AsyncResult<String>> resultHandler) throws IllegalArgumentException {
+      List<String> permissions, Handler<AsyncResult<String>> resultHandler) {
     // Salt and password field should only be set by us.
     if (principal.containsKey(getSaltField()) || principal.containsKey(getPasswordField()))
-      throw new IllegalArgumentException("Salt and password field should not be specified.");
+      Future.failedFuture(new IllegalArgumentException("Salt and password field should not be specified."));
```
pmlopes: shouldn't this be:
```java
if (...) {
  resultHandler.handle(Future.failedFuture(new IllegalArgumentException("Salt and password field should not be specified.")));
  return;
}
```

Otherwise how will you get the exception?

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
```diff
@@ -163,31 +163,31 @@ public void insertUser(String username, String password, List<String> roles, Lis
...
 
     // Shallow copy to prevent details leaking out.
     principal = principal.copy();
 
     if (username != null){
       if (!username.equals(principal.getValue(getUsernameField(), username)))
-        throw new IllegalArgumentException("Conflicting values for username.");
+        Future.failedFuture(new IllegalArgumentException("Conflicting values for username."));
```
pmlopes: same as before...

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
```diff
@@ -163,31 +163,31 @@ public void insertUser(String username, String password, List<String> roles, Lis
...
       principal.put(getUsernameField(), username);
     }
 
     if (roles != null) {
       JsonArray rolesArray = new JsonArray(roles);
       if (!rolesArray.equals(principal.getValue(roleField, rolesArray)))
-        throw new IllegalArgumentException("Conflicting values for roles.");
+        Future.failedFuture(new IllegalArgumentException("Conflicting values for roles."));
```
pmlopes: same as before...

vertx-auth-mongo/src/main/java/io/vertx/ext/auth/mongo/impl/MongoAuthImpl.java
```diff
@@ -163,31 +163,31 @@ public void insertUser(String username, String password, List<String> roles, Lis
...
       principal.put(roleField, rolesArray);
     }
 
     if (permissions != null) {
       JsonArray permissionsArray = new JsonArray(permissions);
       if (!permissionsArray.equals(principal.getValue(permissionField, permissionsArray)))
-        throw new IllegalArgumentException("Conflicting values for permissions.");
+        Future.failedFuture(new IllegalArgumentException("Conflicting values for permissions."));
```
pmlopes: same as before...

pmlopes (changes requested):
We also need some test to cover the failure cases.

DieterDePaepe:
Guess I shouldn't work on this when I'm tired! Fixed and added tests.

pmlopes:
Due to the fact that GDPR is being enforced soon (https://github.com/vert-x3/vertx-auth/pull/184/commits/4f17ba25e22e7ca9b6748a4ae1c25c06b6ece1ca) we should be neutral about personal identifiable data, as the Eclipse Vert.x projects does not want to be liable for any infringements.

For this reason I'll have to close this reject without merging as if it would be merged we would need to implement way more features, like, selective erasure, log erasure + eviction, data exports, etc...

PR #117, opened 2017-01-27
Proper document and polyglot the case of user creation

### LINKED ISSUE #115: Document user creation with auth-jdbc
In Java you can create a secure salt with SecureRandom and then use `io.vertx.ext.auth.jdbc.impl.JDBCAuthImpl#bytesToHex` and `io.vertx.ext.auth.jdbc.impl.JDBCAuthImpl#computeHash` to prepare values for storing a new user.

We should document that.

The problem is that SecureRandom is probably not a code-trans friendly class.

pmlopes:
Fixes #115

### DIFF
--- a/vertx-auth-jdbc/src/main/asciidoc/groovy/index.adoc
+++ b/vertx-auth-jdbc/src/main/asciidoc/groovy/index.adoc
@@ -29,7 +29,7 @@ of those please consult the documentation for the JDBC client.
 
 Once you've got one of those you can create a `link:../../apidocs/io/vertx/ext/auth/jdbc/JDBCAuth.html[JDBCAuth]` instance as follows:
 
-[source,java]
+[source,groovy]
 ----
 import io.vertx.ext.jdbc.JDBCClient
 import io.vertx.ext.auth.jdbc.JDBCAuth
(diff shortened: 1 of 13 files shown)


pmlopes:
There a a couple of things here:

* the hashing call was calling the provider from the strategy (it should be reverse)
* the strategy was not polyglot making it non trivial to use by non java
* missing documentation

This PR addresses the 3 points.

tsegismont (changes requested):
This is great Paulo. I had to figure out that myself when working on the music store demo. So very greatful that this is now documented and polyglot.
However I'm concerned with the random bits generation. Because it can be slow, and even block if there is not enough entropy available.
In the music store, I wrapped this code with executeBlocking (https://git.io/vDJ7a), I believe we should do something similar here.

pmlopes:
What we can do is move:

https://github.com/vert-x3/vertx-web/blob/master/vertx-web/src/main/java/io/vertx/ext/web/sstore/impl/PRNG.java

to core and use it everywhere we need random numbers

vietj:
the goal @tsegismont is not to provide an API for creating secure random for non Vert.x project

vietj:
@pmlopes I would rather move it to this project in auth-common, as it seems more related.

tsegismont:
I didn't suggest that I think. I simply wanted to point out that we
shouldn't let our APIs block the event loop.

2017-01-30 12:27 GMT+01:00 Julien Viet <notifications@github.com>:

tsegismont:
Auth common sounds good to me as well.

2017-01-30 12:29 GMT+01:00 Julien Viet <notifications@github.com>:

pmlopes:
@tsegismont done!

tsegismont:
@pmlopes I can't see how the new commits solve the blocking random gen issue. Did I miss anything? Can we avoid to break the auth factory API?

pmlopes:
The PRNG solves the blocking issue because it wraps the blocking part (generating seed) into an execute blocking call while the rest is not blocking

pmlopes:
The breakage would happen no matter what since there is no vert.x reference in the code and we need one to call the execute blocking

tsegismont (approved):

tsegismont:
LGTM. Please update The 3.4.0 breaking API wiki

PR #118, opened 2017-03-09
added constructor for multiple public keys to JWT

mpluecker:


pmlopes:
Will wait with the merge until 3.4.1 is out.

pmlopes:
@mpluecker I see what you're trying to achieve, what I miss (or I don't get the use case) is that how would a user benefit from this feature when using the `auth-jwt`/`auth-oauth2` modules?

mpluecker:
Actually I do not use the JWT class with the auth-jwt module but in a different context. I want to support public key validation for multiple algorithms (a feature the auth-jwt module would also benefit from). 
I use the vertx-auth package for our internal authentication of microservices. The microservice should only be able to decode the JWT but should not have the private key for signing.

pmlopes:
@mpluecker I understand the usefulness of public key support and for example we use that with both JWT and Oauth2. A use case is to perform AuthN/Z against Keycloak, where keycloak only exports the public key of the server.

Now I can understand that it would be handy to have a constructor where the algorithms could have different public keys but I can't see how is this useful for say `vertx-jwt-auth` or `vertx-oauth2-auth` since these auth handlers do not benefit from this change.

mpluecker:
We do not use vertx-jwt-auth or vertx-oauth2-auth but have our own internal auth-provider which is tailored for our service structure. We use the JWT-Class in this auth provider und need to support multiple algorithms / public keys. Thus we would like to see support for multiple public keys in the JWT class very much.

mpluecker:
Hey, I would appreciate if you could merge this request, since I need it for my work :)

vertx-jwt/src/main/java/io/vertx/ext/auth/jwt/JWT.java
```diff
@@ -124,6 +123,28 @@ public JWT(String key, boolean keyPrivate) {
 
     cryptoMap = Collections.unmodifiableMap(tmp);
   }
+  
+  public JWT(Map<String, PublicKey> publicKeys) {
```
pmlopes, line 127: I'd prefer to see this as a `JsonObject` with a base64 string as the key, the same format we already use in the other constructor, since it is common to get these keys say from google, keycloak, etc... as a string

pmlopes:
@mpluecker the PR #148 includes a more generic way to support this, which supersedes this one.

Issue #105, opened 2016-11-22
add a simple htpasswd provider


pmlopes:
Having a simple htpasswd provider would allow users to use a common file format instead of the custom properties file that will require the usage of apache shiro and blocking API.

This is not a full replacement for properties files since htpasswd only cover AuthN not AuthZ

nevenr:
Hi Paulo,

can you assign this task to me.
Already I did some initial implementation.

Currently implementation is compatible with[ Apache 2.2 htpasswd file formats](https://httpd.apache.org/docs/2.2/misc/password_encryptions.html).
In [Apache 2.4 htpasswd file](https://httpd.apache.org/docs/2.4/misc/password_encryptions.html) [bcrypt](https://en.wikipedia.org/wiki/Bcrypt) format is introduced but I could not find java bcrypt lib that support latest bcrypt algorithm that Apache use. 

Regards,
N.

pmlopes:
@nevenr I can only assign issues to the org members (github limitation).

pmlopes:
However you can make a pull request so we can track that you're working on it.

PR #121, opened 2017-03-13
vertx-auth-htpasswd, add a simple htpasswd provider #105

nevenr:


vertx-auth-htpasswd/src/main/java/io/vertx/ext/auth/htpasswd/impl/HtpasswdAuthImpl.java
```diff
@@ -0,0 +1,110 @@
...
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonObject;
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
+import io.vertx.ext.auth.User;
+import io.vertx.ext.auth.htpasswd.HtpasswdAuth;
+import io.vertx.ext.auth.htpasswd.HtpasswdAuthOptions;
+import org.apache.commons.codec.binary.Base64;
```
pmlopes: Java8 includes a Base64 codec, why rely on an external one?

vertx-auth-htpasswd/src/main/java/io/vertx/ext/auth/htpasswd/HtpasswdAuthOptions.java
```diff
@@ -0,0 +1,79 @@
...
+  private boolean usersAuthorizedForEverything;
+  private String htpasswdFile;
+  private boolean enabledPlainTextPwd;
+
+  public HtpasswdAuthOptions() {
+
+    htpasswdFile = "htpasswd";
+
+    String os = System.getProperty("os.name").toLowerCase();
+    enabledPlainTextPwd = os.startsWith("windows") || os.startsWith("netware");
```
pmlopes, line 23: the `httpd` daemon only accepts plain text on windows or netware but we're not it, why force it also on vert.x side?

vertx-auth-htpasswd/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+    <doc.skip>false</doc.skip>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-auth-common</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>commons-codec</groupId>
```
pmlopes, line 40: We're quite conservative with dependencies, can't we live without this one? `Base64` is part of Java8, SHA/MD5 are just simple `java.security.MessageDigest` instances...

vertx-auth-htpasswd/src/main/java/io/vertx/ext/auth/htpasswd/HtpasswdAuthOptions.java
nevenr: htpasswd file format support five encryption types. Three of them have clear discriminator so we can distinguish them (bcrypt "$2y$", MD5 "$apr1$", SHA1 "{SHA}"). Two of them (CRYPT, PLAIN TEXT) do not have discriminator and they are mutually exclusive. By default this auth implementation will behave like Apache do (on windows and netware plain text is supported, on others crypt), but if one want to change this it can enforced with auth options:
HtpasswdAuthOptions#enablePlainTextAndDisableCryptPwd
or
HtpasswdAuthOptions#enableCryptAndDisablePlainTextPwd

vertx-auth-htpasswd/src/main/java/io/vertx/ext/auth/htpasswd/impl/HtpasswdAuthImpl.java
nevenr: Support for crypt algorithm I found in Apache commons-codec so I used it also for other algorithms.

vertx-auth-htpasswd/pom.xml
nevenr: commons-codec I included for crypt algorithm used in htpasswd file.
If we give up on it, this dependency is not necessary.
Or maybe I can try to extract from commons-codec part related to Crypt algorithm.

nevenr:
For bcrypt I still didn't found java lib that support encryption version used by Apache.
[svenkubiak - jBCrypt](https://github.com/svenkubiak/jBCrypt) or [mindrot - jBCrypt](http://www.mindrot.org/projects/jBCrypt) support older bcrypt version (with discriminator / version identifier "$2a$" ) but Apache use newer  (with "$2y$").

nevenr:
I found code for BCrypt that is up to date. I will add it in soon.

nevenr:
Hi,
Please check now.
If it is OK i can update comments and javadoc.
Regards,
N.

vertx-auth-htpasswd/src/main/java/io/vertx/ext/auth/htpasswd/digest/B64.java
```diff
@@ -0,0 +1,79 @@
...
+     * @param b1
+     *            A byte from the result.
+     * @param b0
+     *            A byte from the result.
+     * @param outLen
+     *            The number of expected output chars.
+     * @param buffer
+     *            Where the output chars is appended to.
+     */
+    static void b64from24bit(final byte b2, final byte b1, final byte b0, final int outLen,
```
pmlopes: Why not use the JDK class?
https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html

vertx-auth-htpasswd/src/main/java/io/vertx/ext/auth/htpasswd/digest/B64.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Generates a string of random chars from the B64T set.
+     *
+     * @param num
+     *            Number of chars to generate.
+     */
+    static String getRandomSalt(final int num) {
+        final StringBuilder saltString = new StringBuilder();
+        for (int i = 1; i <= num; i++) {
+            saltString.append(B64T.charAt(new Random().nextInt(B64T.length())));
```
pmlopes: You create a new random number generator every time, why not just once per method? If the concern is security then use a secure random generator and reseed it would make more sense. Please note that if your random number generator relies on `/dev/random` this is a blocking operation and if there is not enough system entropy you might block the event loop.

pmlopes:
I'm looking at `B64` `getRandomSalt` and `BCrypt` `gensalt` and see 2 places where salt is being generated, they both require a random number generator and in one place we use a simple (non secure) prng `B64` and in the other place a secure prng.

We know by experience (bug reports) that specially on cloud environments your environment may be low on entropy due to virtualization and shared metal, which means that this could block the event loop. It could be interesting if you could use:

https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-common/src/main/java/io/vertx/ext/auth/PRNG.java

Then for example for B64 your code would be something like:

```java
 static String getRandomSalt(final int num) {
  byte[] salt = new byte[num];
  PRNG.nextBytes(salt);
  return java.util.Base64.encoder().encodeToString(salt);
}
```

nevenr:
Hi,

thanks for your time you invested in checking code.

All code related to hashing is wrote in static manner and it would be hard to convert it to non static and than pass through vertex and use io.vertx.ext.auth.PRNG.

But also I think it is not necessary because B64 getRandomSalt and BCrypt gensalt are used for encryption which is not used by HtpasswdAuth.

None of four 'entry point' in io.vertx.ext.auth.htpasswd.digest.Digest:
- bcryptCheck
- md5Check
- shaCheck
- cryptCheck

on execution path do not use doubtful methods. I left it there more to have complete encryption / decryption pair for all algorithms and for testing. All other code is 'package protected' so it should be used.

Regards,
N.

nevenr:
Hi,
I think this is finished, can you please check it.
Thanks.
Regards,
N.

pmlopes:
@nevenr did you sign https://www.eclipse.org/contribute/cla ?

nevenr:
Hi @pmlopes, yes I signed  Eclipse Contributor Agreement.

pmlopes:
@nevenr sorry for the late reply, I've been discussing on merging this work, but since I'm not sure we can merge own crypto algorithms (bcrypt) or we might end up having to do lots of paperwork with eclipse would you mind to refer to this external jar:

http://search.maven.org/#artifactdetails%7Corg.mindrot%7Cjbcrypt%7C0.4%7Cjar

which means that we're not responsible to maintain the crypto code in case of bugs. Also it comes from maven central and the developer is from openbsd so they are way more versed with the code than us...

Sorry for that...

nevenr:
Hi @pmlopes,

unfortunately, as I already mentioned  in [commented on Mar 14](https://github.com/vert-x3/vertx-auth/pull/121#issuecomment-286504508)  [mindrot  jBCrypt 0.4](http://www.mindrot.org/projects/jBCrypt/) is from Jan 2015 and it support older bcrypt version (with discriminator / version identifier "$2a$" ) but Apache use newer (with "$2y$").

Regards,
N.

pmlopes:
Ok but are you aware of any other lib on maven that is up to date, the reason is that we do not want to be responsible for crypto code... I know that when we discussed before we went the other way but I know that if we migrate this module under the eclipse foundation this will be a complicated subject to be approved.

pmlopes:
Alternatively we can compromise by removing the bcrypt part of the pr and merge now and state that it will be added once we clear all the licensing and legal aspects? WDYT?

nevenr:
Let me check what I can find in maven repo for BCrypt.

What about others ( MD5 variant encrypted password / unsalted SHA password /  libc crypt() encoded password )? Should we instead internal implementation (copied code) fall back to Apache commons dependency ?

pmlopes:
Yes, I'm afraid so. I mean let's do that so we can close this now and if later there are no issues with using own algorithm implementations we can always refer to these commits

pmlopes:
Thanks!

PR #124, opened 2017-03-21
Add JWT Auth config options object to provide better typesafety and h…

pmlopes:
…ints to the developer


vertx-auth-jwt/src/main/asciidoc/java/index.adoc
```diff
@@ -131,7 +131,9 @@ case all you need to have is a public key im PEM format.
 
 [source,java]
 ----
-JsonObject config = new JsonObject().put("public-key", "BASE64-ENCODED-PUBLIC_KEY");
+JWTAuthOptions config = new JWTAuthOptions()
+  .setPublicKey("BASE64-ENCODED-PUBLIC_KEY");
+
```
cescoffier: Is this a breaking change ? If so it must be documented and reported to the breaking change list

vertx-auth-jwt/src/main/java/examples/AuthJWTExamples.java
```diff
@@ -29,20 +31,20 @@
...
 
   public void example7(Vertx vertx, String username, String password) {
 
-    JsonObject config = new JsonObject().put("keyStore", new JsonObject()
-        .put("path", "keystore.jceks")
-        .put("type", "jceks")
-        .put("password", "secret"));
+    JWTAuthOptions config = new JWTAuthOptions()
+      .setKeyStore(new JWTKeyStoreOptions()
+        .setPath("keystore.jceks")
```
cescoffier: JCEKS ? is it the right extension ?

vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWTAuthProviderImpl.java
```diff
@@ -36,42 +35,48 @@
...
+        KeyStore ks = KeyStore.getInstance(keyStore.getType());
 
         // synchronize on the class to avoid the case where multiple file accesses will overlap
         synchronized (JWTAuthProviderImpl.class) {
-          final Buffer keystore = vertx.fileSystem().readFileBlocking(keyStore.getString("path"));
+          final Buffer keystore = vertx.fileSystem().readFileBlocking(keyStore.getPath());
 
           try (InputStream in = new ByteArrayInputStream(keystore.getBytes())) {
-            ks.load(in, keyStore.getString("password").toCharArray());
+            ks.load(in, keyStore.getPassword().toCharArray());
```
cescoffier, line 77: This looks like a breaking change for existing auth implementation (I'm not sure it has been implemented by someone else, but maybe it was...)

cescoffier (commented):
I'm not totally sure if it contains breaking change or not.

vertx-auth-jwt/src/main/java/examples/AuthJWTExamples.java
pmlopes: yes, it is different from basic ks, the JCE prefix allows the usage of java cryptographic extensions and read/write more types of data.

pmlopes:
Depends on #148

pmlopes:
I've avoided the breaking change by marking the original `JsonObject` constructor as `GenIgnore` while keeping the new type safe method using the same structure.

mslinn:
`vertx-examples/web-examples/src/main/java/io/vertx/example/web/authorisation/Server.java` should be updated:
```
JWTAuth jwt = JWTAuth.create(vertx, new JsonObject()
        .put("keyStore", new JsonObject()
            .put("type", "jceks")
            .put("path", "keystore.jceks")
            .put("password", "secret")));
```

pmlopes:
@mslinn the old format is still acceptable but deprecated but indeed the examples should be updated for the current API

mslinn:
People like me use the sample apps for guidance. I would like to follow current recommended practices, but that is only possible when they are demonstrated.

PR #131, opened 2017-04-06
Context secure PRNG + docs

pmlopes:
This is a proposal for a vertx context secure PRNG similar to Java's ThreadLocalRandom.

The rationale behind this is that random numbers are used in several places in the project, e.g.:

* Mongo Auth
* Mail Client
* Vertx-web CSRF
* Vertx-web htdigest
* SockJS Base Transport
* JDBC Auth
* Web Session store's
* Web Session
* plus users might also use random numbers in their apps

We know that SecureRandom is a blocking API if there is not enough entropy, so having a PRNG for each of these handlers will consume some entropy plus resources. Using this alternative approach there will be a max of 1 PRNG per vert.x context.

### DIFF
--- a/vertx-auth-common/src/main/java/io/vertx/ext/auth/VertxContextRandom.java
+++ b/vertx-auth-common/src/main/java/io/vertx/ext/auth/VertxContextRandom.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2015 Red Hat, Inc.
+ *
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v1.0
+ *  and Apache License v2.0 which accompanies this distribution.
+ *
+ *  The Eclipse Public License is available at
+ *  http://www.eclipse.org/legal/epl-v10.html
+ *
+ *  The Apache License v2.0 is available at
+ *  http://www.opensource.org/licenses/apache2.0.php
+ *
+ *  You may elect to redistribute this code under either of these licenses.
+ */
+package io.vertx.ext.auth;
+
+import io.vertx.codegen.annotations.GenIgnore;
+import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.core.Context;
+import io.vertx.core.Vertx;
+
+import java.util.Base64;
+
+/**
+ *
+ * A secure non blocking random number generator isolated to the current context. The PRNG is bound to the vert.x
+ * context and setup to close when the context shuts down.
+ *
+ * When applicable, use of VertxContextRandom rather than create new PRNG objects is helpful to keep the system entropy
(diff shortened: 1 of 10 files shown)


vertx-auth-common/src/main/java/io/vertx/ext/auth/VertxContextRandom.java
```diff
@@ -0,0 +1,140 @@
...
+    final Context currentContext = Vertx.currentContext();
+    if (currentContext != null) {
+      return current(currentContext);
+    }
+
+    throw new IllegalStateException("Not running in a Vert.x Context.");
+  }
+
+  /**
+   * Get or create a secure non blocking random number generator using the current vert.x context. This method will not
```
tsegismont: `current` / `provided`

vertx-auth-common/src/main/java/io/vertx/ext/auth/VertxContextRandom.java
```diff
@@ -0,0 +1,140 @@
...
+   * throw an exception.
+   *
+   * @param context a Vert.x context.
+   * @return A secure non blocking random number generator.
+   */
+  static VertxContextRandom current(final Context context) {
+
+    // attempt to load a PRNG from the current context
+    PRNG random = context.get(VertxContextRandom.class.getName());
```
tsegismont: Why not use \_\__.vertx.sthg_ as with do with other internal stuff?

vertx-auth-common/src/main/java/io/vertx/ext/auth/VertxContextRandom.java
```diff
@@ -0,0 +1,140 @@
...
+ * <p>
+ * When applicable, use of VertxContextRandom rather than create new PRNG objects is helpful to keep the system entropy
+ * usage to the minimum avoiding potential blocking across the application.
+ * <p>
+ * The use of VertxContextRandom is particularly appropriate when multiple handlers use random numbers.
+ *
+ * @author <a href="mailto:plopes@redhat.com">Paulo Lopes</a>
+ */
+@VertxGen
+public interface VertxContextRandom {
```
tsegismont: Call it PRNG not Random. Otherwise it sound like a simple ThreadLocalRandom

tsegismont (changes requested):

tsegismont:
> plus users might also use random numbers in their apps

I wouldn't make this a general purpose recommendation. If users need a simple random they should use the tools available in their language.

Could you please show an example of Javascript code which would benefit from this change?

pmlopes:
The asciidocs now include an example for all generated languages. It shows that one can generate random Strings and Ints using the same API for all languages.

tsegismont (approved):

PR #135, opened 2017-04-12
Fixes #129: rename package, should not be breaking change since it's …

### LINKED ISSUE #129: Both modules vertx-jwt and vertx-auth-jwt have the same package name io.vertx.ext.auth.jwt
### Version

* vert.x core 3.4.x:
* vert.x auth 3.4.x:

### Context

Both modules  **vertx-jwt** and **vertx-auth-jwt** have the same package name **io.vertx.ext.auth.jwt**. Conflict leads to runtime error:
`java.lang.NoClassDefFoundError: io/vertx/ext/auth/jwt/JWTAuth`
 in osgi environment.

pmlopes:
…a internal API not public facing


pmlopes:
@cescoffier @vietj although this looks like an breaking change (rename of package) the module in question is a "internal" helper for the public facing modules "jwt-auth" and "oauth2-auth", would this "fix" be ok?

PR #136, opened 2017-04-13
Fixes #133

### LINKED ISSUE #133: jdbc-auth default hashing strategy is weak
Current algorithm just does 1x SHA256 + salt which is unsecure, 

see: https://blog.mozilla.org/security/2011/05/10/sha-512-w-per-user-salts-is-not-enough/

consider a stronger solution such as:

https://github.com/defuse/password-hashing

pmlopes:
This fix is 100% backwards compatible as shown in test.

The choosen approach is to use a application level configuration of nonces that is **NOT** expected to be stored in the database, this way we present precomputed hash attacks and rainbow table attacks, leaving room only for brute force attacks.

### DIFF
--- a/vertx-auth-jdbc/src/main/asciidoc/groovy/index.adoc
+++ b/vertx-auth-jdbc/src/main/asciidoc/groovy/index.adoc
@@ -116,6 +116,55 @@ conn.updateWithParams("INSERT INTO user VALUES (?, ?, ?)", [
 
 ----
 
+WARNING: Hashing user password with salt can be not enough, this approach his good enough for avoiding rainbow tables
+attacks or precomputed table attacks but if the attacker gets the database it will be easier to setup a brute force
+attack. This kind of attack is slower but all required information is given: the hash and the salt.
+
+To make the hash attack more complex the default strategy allows you to provide an application level list of nonces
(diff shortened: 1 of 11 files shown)


jponge (approved):

kordian2k:
Hi Paulo,

Thanks for the magnificent work on vertx! I have two questions w.r.t. this PR:

When comparing the password against the hash using
` if (hashedStoredPwd.equals(hashedPassword)) {`
the equals method returns after the first character mismatch. Isn't this vulnerable to timing attacks?

Why don't you simply use BCrypt to hash passwords?

Cheers!
K.

pmlopes:
The stored and hash variables are the data from the database (static) and the result of the hashing algorithm. Since Java strings are immutable by nature unless you're being attacked by someone modifying your JVM process RAM you should not consider it a risk.

If someone does control your JVM process RAM then you have more serious issues than timing attacks.

pmlopes:
Also bcrypt would not solve anything since it's a hashing algorithm like the one we use. We can override the current default with bcrypt as long as there's a proper JVM implementation.

The algorithm we choosed as default still remains as default for the reason of backwards compatibility but nothing stops users to extend the default.

PR #148, opened 2017-06-29
Fixes #147: properly convert from DER to JWS signatures when dealing …

### LINKED ISSUE #147: JWT verification problem when using elliptic curve cryptography
### Version

* vert.x core: 3.4.2
* vert.x auth: 3.4.2

### Context
I encountered a verification problem w/ JWT tokens generated using Vert.x and verified using the python package [PyJWT](https://github.com/jpadilla/pyjwt). Generating token w/ PyJWT and verifying using Vert.x leads to a similar error.

The problem only exists when using elliptic curve cryptography (`ES512`). 

### Do you have a reproducer?

* https://github.com/kordian2k/jwt_elliptic_error

### Steps to reproduce
The reproducer constructs a *docker image* based on Debian Jesse. Prerequisites are a working docker installation. Moreover, the user running docker must have rights to run wo/ using `sudo` (e.g. by adding the user to the docker group).

1. `mvn package`
2. `docker run -it <image_id>`

### Extra
I understand the problems when using ECC w/ OpenJDK on which the docker image is based. However, a similar image running on **Hotspot** does not change behavior.

I do not know if the problem comes from Vert.x or PyJWT. However, it appears PyJWT was it least tested against a [Ruby implementation](https://github.com/jpadilla/pyjwt/issues/168). In general it would be desirable to have a compatible implementation.

pmlopes:
…with ECDSA

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ src/main/resources/ext-js/*.js
 src/main/java/io/vertx/java/**/*.java
 src/main/groovy/io/vertx/groovy/**/*.groovy
 *.swp
+node_modules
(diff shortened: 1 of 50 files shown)


pmlopes:
The code is verified to be working by @kordian2k

pmlopes:
I will go ahead and merge this since it has been tested by the community and I need to start testing all dependencies `vertx-web`

PR #151, opened 2017-07-26
Adding bcrypt support from 'org.mindrot:jbcrypt:0.4'

nevenr:
- unfortunately 'org.mindrot:jbcrypt:0.4' support only older bcrypt version (prefix "$2a$")
- current version of apache htpasswd use newer bcrypt (prefix "$2y$") and it is not currently supported with vertx-auth-htpasswd


nevenr:
I could not find any lib in maven repo that support bcrypt "$2y$" version.
Then I thought it is better to support older version than nothing.

pmlopes:
Thanks @nevenr, there is now support to add custom hash algorithms outside the JDK and outside the project umbrella. Since jbcrypt is not up to date I'll close this PR and further enhancements can be added later on.

PR #152, opened 2017-07-31
add a configuration option to use HMac secrets without a keystore

inve1:
You added a way to pass in arbitrary secrets to HMAC on the JWT object recently, but there's no way to use that in JwtAuth yet. Adding a configuration hook for that


pmlopes (changes requested):
There are a couple of comments and we miss at least a basic test to verify the functionality.

inve1:
fixed comments, take another look

pmlopes (approved):

Issue #153, opened 2017-08-07
AccessToken.refresh error


johnoliver:
### Version

* vert.x auth: 3.4.2

### Context

When calling AccessToken.refresh I get the following error:

```
Cannot supply multiple client credentials. Use one of the following: credentials in the Authorization header, credentials in the post body, or a client_assertion in the post body
```

Seems like `AccessTokenImpl.refresh` needs to add `put("authorizationHeaderOnly", true)` to the api query, similar to the `AccessTokenImpl.introspect` method. 

Would adding this break other applications?

pmlopes:
I don't believe it will break anything.

PR #154, opened 2017-08-08
Pass authorizationHeaderOnly=true to AccessToken revoke and refresh calls

johnoliver:
As discussed in #153 auth in both headers and body causes an error, so add authorizationHeaderOnly=true similar to the introspect method.


pmlopes:
I think we need to partially revert this:

https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/intro_understanding_refresh_token_oauth.htm#!

refresh does need client id, client secret

PR #158, opened 2017-09-01
Implemented leeway as stated on the RFC to allow validation on clock …

### LINKED ISSUE #157: JWT-Auth "now" is truncating instead of rounding.
### Version

* vert.x core: 3.4.2
* vert.x auth: 3.4.2

### Context

I've noticed that when using Auth-JWT, sometimes the validation of the token fails because iat is greater than now (Current time in seconds). Everything else passes, but iat. I've noticed this behavior with Google's OAuth Implementation.

In testing I have deduced that Google is rounding their timestamp, while Java is truncating it. This results in Java's Timestamp sometimes being behind Google's.

Example:

Now in milliseconds (Obtained via `System.currentTimeMillis()`)
1504218035724

Google: 1504221636
Vertx: 1504218035

### Current Code Location

Vertx Truncates the time here:
https://github.com/vert-x3/vertx-auth/blob/8e5c0f5b42cd7531f864eae974a7aa42230c33d7/vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWTAuthProviderImpl.java#L117

### Steps to reproduce 
(Unfortunately I'm not able to provide the code for reproduction. Company Property and all that.)

1. Do OAuth2 with Google to spec to get Token
2. Output the Token's decoded Payload
3. Output `System.currentTimeMillis()`
4. Compare the Payload's iat with current time.

### Suggested Fix
Round the time based on normal conventions (5 and up rounds up, 4 and under rounds down.)

pmlopes:
…drifting

Fixes #157

Reference: https://www.rfc-editor.org/rfc/rfc7519.txt

### DIFF
--- a/vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWTAuthProviderImpl.java
+++ b/vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWTAuthProviderImpl.java
@@ -116,10 +116,10 @@ public class JWTAuthProviderImpl implements JWTAuth {
       // All dates in JWT are of type NumericDate
       // a NumericDate is: numeric value representing the number of seconds from 1970-01-01T00:00:00Z UTC until
       // the specified UTC date/time, ignoring leap seconds
-      final long now = (System.currentTimeMillis() / 1000) + leeway;
+      final long now = (System.currentTimeMillis() / 1000);
 
       if (payload.containsKey("exp") && !ignoreExpiration) {
-        if (now >= payload.getLong("exp")) {
+        if (now - leeway >= payload.getLong("exp")) {
           resultHandler.handle(Future.failedFuture("Expired JWT token: exp <= now"));
           return;
         }
@@ -128,7 +128,7 @@ public class JWTAuthProviderImpl implements JWTAuth {
       if (payload.containsKey("iat")) {
         Long iat = payload.getLong("iat");
         // issue at must be in the past
-        if (iat > now) {
+        if (iat > now + leeway) {
           resultHandler.handle(Future.failedFuture("Invalid JWT token: iat > now"));
           return;
         }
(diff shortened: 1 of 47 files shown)


vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWTAuthProviderImpl.java
```diff
@@ -114,7 +116,7 @@ public void authenticate(JsonObject authInfo, Handler<AsyncResult<User>> resultH
       // All dates in JWT are of type NumericDate
       // a NumericDate is: numeric value representing the number of seconds from 1970-01-01T00:00:00Z UTC until
       // the specified UTC date/time, ignoring leap seconds
-      final long now = System.currentTimeMillis() / 1000;
+      final long now = (System.currentTimeMillis() / 1000) + leeway;
```
jwagnerLegacy: Related to Issue #157 & PR #158

Shouldn't we account for possible drift in both directions? + or - leeway?

Example:

```Java
final long now_max = (System.currentTimeMillis() / 1000) + leeway;
final long now_min = (System.currentTimeMillis() / 1000) - leeway;
```

Then, at line 130:

```Java
if (iat > now_min || iat > now_max) {
      resultHandler.handle(Future.failedFuture("Invalid JWT token: iat > now"));
      return;
 }
```

Another option would a a `positiveLeeWay` variable and a `negactLeeWay `variable to be set. and applied respectively.

pmlopes:
Adding a negative leeway would mean that we're are behind in time. If that happens the value of now is always in the past so validation of issued at, not before will always fail and it probably will only affect the expiration validation. And in that case I am not sure if that's a desired behavior.

pmlopes:
But I see that this fix is not correct, for iat and nbf it should be validated against now + leeway, and for exp it should be now - leeway

PR #162, opened 2017-10-05
fix: NPE when calling JWTOptions.getNoTimestamp()

SgrAlpha:
The problem is simple, when you create a blank JWTOptions or from JsonObject without noTimestamp field, and then call getNoTimestamp(), it will throw a NPE, because JsonObject. getBoolean() will return NULL, but the return type is boolean not Boolean.


pmlopes:
Thanks!

PR #175, opened 2018-01-12
Google uses certificates as input for the JWT engine

pmlopes:
Using the official google certificates as input + testing


pmlopes:
This is a trivial fix (so merged right away)

PR #179, opened 2018-01-17
Implement a hash strategy based on PBKDF2

### LINKED ISSUE #178: vertx-auth doesn't use current OWASP password hashing recommendation
This is basically a followup to #133. 

The addition of what you call "nonces" to the salted SHA512 is not what the state of the art in cryptographic password hashing is. You want to use either bcrypt (https://github.com/jeremyh/jBCrypt) or the newer argon2 (https://github.com/P-H-C/phc-winner-argon2). Java bindings for argon2 are available.

The problem with just using a single round of SHA512 is that it can still be computed *very quickly*. Someone with access to a modern GPU like a GTX 1080 can very easily try *millions* of passwords per second. The point of functions like argon2 and bcrypt is to make such attacks harder to compute: A single password attempt actually takes a configurable amount of memory and CPU time to compute. That way, even a very sophisticated attacker can only run maybe a few thousand (or even less, with higher parameters) passwords per second. Adding these nonces is mostly just spreading out the problem; if an attacker acquires the nonces, then they can still mount the brute force attack as easily as without the nonces. And since in most setups, the database and the nonces will sit on the same server, a compromise of one of the two is almost always a compromise of both anyways.

Now, you might think "what if we encrypt the nonces?". But that is an equally bad idea: Now where do you store the encryption key for the nonces? So the solution to this issue is to actually migrate to a real password hashing algorithm.

pmlopes:
Fixes #178 by following the updated OWASP recommendation:

https://www.owasp.org/index.php/Password_Storage_Cheat_Sheet


yawkat:
This does not fix the API issues highlighted in #178. It does not offer any way of rehashing, for example.

sonOfRa:
It may also be a good idea to use PBKDF2-HMAC-SHA512 instead of SHA1. It's not that SHA1 is broken *in the context of PBKDF*, but SHA512 is a factor of 8 slower than SHA1 on a GTX 1080 hashcat implementation. This slowness is a desirable property.

For example, for the plain SHA512 in the old default, I could calculate about a billion hashes per second. With PBKDF2-HMAC-SHA1 that number drops to around 3 *million* hashes per second. PBKDF-2HMAC-SHA512 drops this by another factor of 8 to around 400k hashes per second. In comparison, functions like bcrypt (which you should seriously consider) drop that another factor of *30* to around 13k hashes per second. I don't have any updated benchmarks for argon2, but that should drop it by another factor of around 2 to 3.

source for hashing benchmarks: https://gist.github.com/epixoip/a83d38f412b4737e99bbef804a270c40

pmlopes:
@sonOfRa I totally agree that we should pull bCrypt but please bear with me for a moment. Doing so is not as trivial as one might think. From a developer perspective, it's just a dependency away, however since it is a crypto dependency it will most likely require approval from the Eclipse Foundation and several legal iterations need happen before it all gets clear.

It is however possible and probably a nice open source project to create a sub project outside the realm of this project that provides even more secure hashing strategies, e.g.: a project you host under your github name and you can add `bCrypt` or indeed the `argon2` algorithm.

pmlopes:
@yawkat rehashing was never a concern from this module and it can be implemented externally as documented on sites such as: https://veggiespam.com/painless-password-hash-upgrades/

yawkat:
Following the methodology of that article would require a user to basically reimplement your JDBCAuthImpl.

There is no good reason to not add *some* support for such basic functionality.

pmlopes:
@yawkat lets be realistic here, adding a extra column for a "must re-hash" would not affect the code, you could after the authN occurs, verify if the users has the flag active (again without changing your code) and you could iterate over a list of strategies to re-encode.

True this would be some code, but we have to agree that it's not a bug, **BUT** a great new feature!

Now, we're in the middle of a bug release from 3.5.0 to 3.5.1 which means we can't break API unless it is utterly broken (which isn't) **BUT** we can proper plan this as a new feature for 3.6, wdyt?

pmlopes:
@yawkat @sonOfRa I'd like to invite you guys help spec'ing and coding (if you would like) the issue #180

PR #182, opened 2018-01-25
Allow crypto providers outside the JDK

pmlopes:
This is a follow up of https://github.com/vert-x3/vertx-auth/issues/180


pmlopes:
Given the fact that `htpasswd` also relies on several hashing algorithms this PoC should be extended to allow it to be used there. This would mean that the missing `bCrypt` algorithm could be added be 3rd party libraries, avoiding the complex legal review from the organization with regards to cryptographic code.

pmlopes:
Using this approach we can add any hashing algorithm we need, as a PoC I've added:

* SHA512
* PKDBF2

As the strategies for mongo and JDBC auth. To prove this works for non JDK crypto libraries I've patched the htpasswd auth to implement:

* Plaintext
* Crypt(3)
* APR1
* SHA1

using apache common codec dependency.

vertx-auth-common/src/main/java/io/vertx/ext/auth/HashString.java
```diff
@@ -0,0 +1,131 @@
+package io.vertx.ext.auth;
```
jponge, line 16: Header missing ;-)

vertx-auth-common/src/main/java/io/vertx/ext/auth/HashString.java
```diff
@@ -0,0 +1,131 @@
...
+      encoded = encoded.replaceAll("\\{", "\\$\\{");
+      encoded = encoded.replaceAll("\\}", "\\}\\$");
+      if (encoded.charAt(0) != '$') {
+        encoded = "$$" + encoded;
+      }
+    }
+
+    parts = encoded.split("\\$");
+
+    switch (parts.length) {
```
jponge: It doesn't seem to me like there is anything wrong here, but shouldn't you still have a `default`, _just in case_? (and thrown an `IllegalStateException` or something?)

vertx-auth-common/src/main/java/io/vertx/ext/auth/HashingAlgorithm.java
```diff
@@ -0,0 +1,43 @@
+package io.vertx.ext.auth;
```
jponge, line 16: Headers, headers, I'm a header maniac 😉

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/hash/PBKDF2.java
```diff
@@ -0,0 +1,74 @@
+package io.vertx.ext.auth.impl.hash;
```
jponge, line 16: ⚖️

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/hash/PBKDF2.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+
+  @Override
+  public String id() {
+    return "pbkdf2";
+  }
+
+  @Override
+  public Set<String> params() {
```
jponge, line 59: This is where we'd love having immutable collections in Java 8.

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/hash/SHA512.java
```diff
@@ -0,0 +1,34 @@
+package io.vertx.ext.auth.impl.hash;
```
jponge, line 16: ⚖️

vertx-auth-htpasswd/src/main/java/io/vertx/ext/auth/htpasswd/HtpasswdAuthOptions.java
```diff
@@ -35,18 +32,16 @@ public HtpasswdAuthOptions(HtpasswdAuthOptions that) {
...
+  public HtpasswdAuthOptions setEnablePlainText(boolean enabledPlainText) {
+    this.enabledPlainText = enabledPlainText;
     return this;
-
   }
 
-  public HtpasswdAuthOptions enableCryptAndDisablePlainTextPwd() {
-    enabledPlainTextPwd = false;
-    return this;
+  public boolean isEnablePlainText() {
```
jponge: `isPlainTextEnabled()`?

vertx-auth-htpasswd/src/main/java/io/vertx/ext/auth/htpasswd/impl/hash/APR1.java
```diff
@@ -0,0 +1,21 @@
+package io.vertx.ext.auth.htpasswd.impl.hash;
```
jponge, line 16: ⚖️

jponge (commented):
Minor comments added

vertx-auth-common/src/main/java/io/vertx/ext/auth/HashString.java
```diff
@@ -0,0 +1,131 @@
...
+  public HashString(String id, Map<String, String> params, String salt) {
+    this.id = id;
+    this.params = params;
+    this.salt = salt;
+  }
+
+  public HashString(String encoded) {
+    String[] parts;
+
+    if (encoded.charAt(0) != '$') {
```
vietj: check string is not empty

vertx-auth-common/src/main/java/io/vertx/ext/auth/HashString.java
```diff
@@ -0,0 +1,131 @@
...
+        hash = parts[4];
+        break;
+    }
+  }
+
+  public String id() {
+    return id;
+  }
+
+  public String param(String param) {
```
vietj, line 92: null params is considered as caller error and should not return null

vertx-auth-common/src/main/java/io/vertx/ext/auth/HashingAlgorithm.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.Base64;
+import java.util.Collections;
+import java.util.Set;
+
+public interface HashingAlgorithm {
+
+  Base64.Decoder B64DEC = Base64.getDecoder();
```
vietj: what's the purpose to keep such constant fields ?

vertx-auth-common/src/main/java/io/vertx/ext/auth/HashingStrategy.java
```diff
@@ -0,0 +1,28 @@
+package io.vertx.ext.auth;
+
+import io.vertx.ext.auth.impl.HashingStrategyImpl;
+
+import java.util.Map;
+import java.util.ServiceLoader;
+
+public interface HashingStrategy {
```
vietj: javadoc

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/hash/PBKDF2.java
```diff
@@ -0,0 +1,74 @@
...
+import java.security.spec.InvalidKeySpecException;
+import java.util.HashSet;
+import java.util.Set;
+
+public class PBKDF2 implements HashingAlgorithm {
+
+  private static final int DEFAULT_ITERATIONS = 10000;
+
+  private static final Set<String> DEFAULT_CONFIG = new HashSet<String>() {{
+    add("it");
```
vietj: rather Collections.singleton("id"), that will create an anonymous inner class

vertx-auth-common/src/main/java/io/vertx/ext/auth/HashingAlgorithm.java
vietj: move them to impl classes where it's actually used

vertx-auth-htpasswd/src/test/java/io/vertx/ext/auth/htpasswd/EncodingTest.java
```diff
@@ -0,0 +1,32 @@
+package io.vertx.ext.auth.htpasswd;
+
+import io.vertx.ext.auth.HashingStrategy;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class EncodingTest {
```
vietj, line 8: need to test behavior with malformed input

vertx-auth-htpasswd/src/main/java/io/vertx/ext/auth/htpasswd/impl/hash/Crypt.java
```diff
@@ -0,0 +1,21 @@
+package io.vertx.ext.auth.htpasswd.impl.hash;
+
+import io.vertx.ext.auth.HashString;
+import io.vertx.ext.auth.HashingAlgorithm;
+
+import org.apache.commons.codec.digest.UnixCrypt;
```
vietj, line 21: hhmm. apache commons hell... can we avoid this ?

pmlopes: The dependency is optional so it will cause a class not found exception, which in turn  will make this provider not being available to the service loader.

This specific provider is needed for the htpasswd handler though.

vertx-auth-common/src/main/java/io/vertx/ext/auth/HashingAlgorithm.java
pmlopes: Not really,  for example htpasswd module defines a custom provider for apache specific algorithms

PR #193, opened 2018-03-02
Added support for generating random ints with a bound

jessym:
I want to use the Vert.x managed `Random` for picking a random element from a list.

Because it's not trivial to calculate `randomInt(int bound)` from `randomInt()`, I made this PR.


pmlopes:
:+1:

PR #199, opened 2018-03-13
Enable access to the RAW token field

julienBaleizao:
In some cases, accessing raw token could be useful.


pmlopes:
This is a duplicate of the `principal()` call which is a common API for all auth.

pmlopes:
This would add a duplicate method to existing functionality

PR #200, opened 2018-03-14
Fixes #197

### LINKED ISSUE #197: Get user information failed with Facebook
### Version

* vert.x core: 3.5.1
* vert.x auth: 3.5.1

### Context

When I try to get the user information following this code:
`(OAuth2TokenImpl oAuth2Token).userInfo(...);`

The future failed with the message "Cannot handle Content-Type: text/javascript"
After looking at the Vertx code I see that you only handle the content type application/json, application/x-www-form-urlencoded or text/plain. The problem is that Facebook responds with content type text/javascript for the user information.
Do you have planned to fix this issue ? Is it the right way to retrieve user information ?

### Steps to reproduce

1. Use the FacebookAuth Vertx provider to authenticate an user
2. Once the Token is received call the method userInfo

pmlopes:
Specify that we prefer JSON responses for userInfo requests


pmlopes:
Trivial fix (no side effects or API breakage)

PR #209, opened 2018-04-19
remove duplicate checks in token introspection

### LINKED ISSUE #208: OAuth2TokenImpl: duplicate checks in token introspection
### Version

* vert.x core: master
* vert.x auth: master

### Context

https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2TokenImpl.java#L519  + ff 

Both the check for the `active` flag and the correct `client_id` are duplicate in the code.   The check for `client_id` is done slightly different, but functionally identical (looking into a local variable instead of getting the provider from scratch). 

### Do you have a reproducer?

Not a functionality bug, "just" housekeeping

### Steps to reproduce

n.a.

### Extra

n.a.

nkuehn:
fixes #208 

Just housekeeping.  
Feel free to do it yourself in a local branch - it's trivial to do from scratch.


pmlopes:
There has been a large refactoring in this code, I believe the refactor already fixed this, if you still spot issues please reopen.

PR #218, opened 2018-05-30
Fixes #165

### LINKED ISSUE #165: Multiple authentication methods used
### Version

* vert.x core: 3.5.0
* vert.x auth: 3.5.0

### Context

Since #154 was partially revoked, it seems as though client_id/client_secret will be passed in the body of the requests and the client password will be added to the Authorization header. Comments say that Salesforce requires this (https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/AccessTokenImpl.java#L211), according to [their docs](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/intro_understanding_refresh_token_oauth.htm#) it seems like `client_id` is required, but does say client_secret is optional. 

If Salesforce does require this it is unfortunate as according to the standard:

```
The client MUST NOT use more than one authentication method in each request.
```
https://tools.ietf.org/html/rfc6749#section-2.3

So currently any service that implements the standard correctly will currently reject these requests. Probably need to add a configuration parameter to support both.

jakefeasel:
This omits the client_secret parameter from the form body when basic authorization is used.


pmlopes:
Thanks!

PR #227, opened 2018-07-02
Fixes #226

### LINKED ISSUE #226: JWTAuth does not expose a way to configure JWKs
This should be present as it should become the expected way to handle keys

pmlopes:


pmlopes:
Trivial fix (already thoroughly tested on vertx-jwt)

PR #229, opened 2018-07-02
Fixes #228

### LINKED ISSUE #228: Deprecate JWT key loading from keystores
In the current implementation the JWT code can read both PEM file format files or get the keys from a java keystore. The java keystore is not a standard but a implementation detail inherited when the original code was ported from Yoke to Vert.x3.

The code assumes that keys are named according to the algorithms expected by the JWT format and makes even more assumptions for RSA keys where signature algorithms are derived from the certificate file.

In real life, Google, Keycloak, etc... use either PEM files or JWKs.

Many users have issues with this self made keystores so (since it was a bad design decision from my side many years ago on Yoke) I'd like to mark the usage of keystores deprecated.

pmlopes:


pmlopes:
As discussed here: https://groups.google.com/forum/#!topic/vertx-dev/SHXCvf-lass

PR #230, opened 2018-07-06
In OAuth, RP stands for Relying Party

BrunoVernay:
"Relay Party" appears in a lot of files. You may have some more effective mean to perform a "change-all".
(If you agree on the change https://en.wikipedia.org/wiki/Relying_party )

By the way, what would be a good place to discuss OAuth documentation (I have a few remarks). vertx-dev Google group?


pmlopes (approved):

pmlopes:
@BrunoVernay we can discuss on gitter or the Google group. Thanks!

Issue #231, opened 2018-07-12
In Auth2: logs more info, debug  messages to help debugging and understanding


BrunoVernay:
### Version

* vert.x core: 3.5.2
* vert.x auth: 3.5.2

### Context

I try to use OAuth (OpenID Connect) to authenticate to Keycloak, but have a hard time figuring out what happen and what the API does.

It would be nice to know what is exchanged with the "OpenID Provider" or "Authorization Server", front and especially  **back** channel. 
Also know the content and state (expired?) of the tokens and the state of the flow (have Authorization Code, has Access Token, has IDToken ...)

### Do you have a reproducer?

* https://gitlab.com/BrunoVernay/oauth-vertx-kotlin

BrunoVernay:
Also some related pieces are in https://github.com/vert-x3/vertx-web/blob/master/vertx-web/src/main/java/io/vertx/ext/web/handler/impl/OAuth2AuthHandlerImpl.java

pmlopes:
Code has been merged.

PR #232, opened 2018-07-13
Add logs info/debug

BrunoVernay:
and a tiny comment change
Trying to fix issue #231

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2ResponseImpl.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2ResponseImpl.java
@@ -18,18 +18,6 @@ public class OAuth2ResponseImpl implements OAuth2Response {
 
   public OAuth2ResponseImpl(int statusCode, MultiMap headers, Buffer body) {
     LOG.info("New response: statusCode: "+ statusCode );
-    if (LOG.isDebugEnabled()) {
-      final JsonObject json = new JsonObject();
-      final JsonObject jsonHeaders = new JsonObject();
-      if (null != headers) {
-        headers.forEach((entry) -> jsonHeaders.put(entry.getKey(), entry.getValue()));
-      }
-      json.put("headers", jsonHeaders);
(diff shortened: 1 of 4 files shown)


vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2ResponseImpl.java
```diff
@@ -4,16 +4,32 @@
...
 public class OAuth2ResponseImpl implements OAuth2Response {
 
+  private static final Logger LOG = LoggerFactory.getLogger(OAuth2ResponseImpl.class);
   private final int statusCode;
   private final MultiMap headers;
   private final Buffer body;
 
 
   public OAuth2ResponseImpl(int statusCode, MultiMap headers, Buffer body) {
+    LOG.info("New response: statusCode: "+ statusCode );
```
pmlopes, line 20: This is a touchy subject, if we log everything we might end up being marked as a security vulnerability as secure tokens get "leaked" into the logs, in order to inspect the response we need a interceptor like api to allow custom code to inspect the response and it's up to the user code to decide if they want to dump it to a log file, not up to the framework.

BrunoVernay: I was hoping that if user sets DEBUG level it would be OK to dump everything, but I understand.
(In the highlighted line, the `statusCode` is just the HTTP response status code, right? )

BrunoVernay (commented):
Removed sensible data dumps in logs

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2ResponseImpl.java
pmlopes: status codes are fine, it's more of the body itself, i think headers could be fine too, not sure about `Authorization` as it is a token in transit though

recena (approved):

pmlopes:
Thanks!

PR #256, opened 2019-01-23
3.6

Ngosti2000:
vertx-auth-common does not export io.vertx.ext.auth.impl yet its required by ouath2.OSGi container Karaf hence complains when trying to resolve vertx-auth-oauth2 bundle in karaf.


pmlopes:
This PR is not valid.

PR #257, opened 2019-01-23
enable export of impl package in vertx-auth-common for OSGI containers

Ngosti2000:
vertx-auth-oauth2 requires impl package from vertx-auth-common which apparently is not exported hence Karaf is unable to resolve the outh2 bundle.


pmlopes:
OSGi metadata has been removed and will not be supported on 4.0

PR #261, opened 2019-02-08
reuse `HttpClient` between oauth requests

jo5ef:
Currently, a new `HttpClient` is initialized and closed again for *every* oauth request. This causes a new connection to be established for *every request* which can have a significant impact on performance (see [this vertx-group conversation](https://groups.google.com/forum/#!msg/vertx/IrC5pFqj1J8/E51_8MH-GgAJ)). This PR changes the `OAuth2API` class to keep a reference to the `HttpClient` that is re-used between requests.


pmlopes:
:+1: Thanks!

PR #263, opened 2019-02-22
Fix copy paste comments in JWT examples

humb1t:
During copy paste of some examples 
```
// In this case we are forcing the provider to ignore the `exp` field
```
comment was used three times. This fix rewrites two of them to align with the code samples.


pmlopes:
:+1: thanks!

PR #278, opened 2019-03-26
Issue 277 - Disable validation of refresh token

vinscom:
To avoid any breaking changes. I have only disabled validation of refresh token. Means, decoded refresh token is still available, but, signature is not validated.


pmlopes:
Thanks!

PR #290, opened 2019-09-04
Allow chain auth to do 'AND' and 'OR' operations

### LINKED ISSUE #183: ChainAuthHandlerImpl always authorizes authenticated users
### Version

* vert.x core: 3.5.0
* vert.x auth: 3.5.0

### Context

The chaining of multiple auth mechanisms does not work as expected when an user provides correct credentials, but does not have the authorities to access a resource.
 
The io.vertx.ext.web.handler.impl.ChainAuthHandlerImpl class overrides the addAuthority and addAuthorities methods from the AuthHandlerImpl class to delegate to its chained handlers, but does not override the AuthHandlerImpl.authorize method, which uses the empty authorities field of the ChainAuthHandlerImpl class. Since the authorities field is always empty, the authorize method always succeeds and the user is allowed to access the resource.
Another issue with the ChainAuthHandlerImpl is that it does not support the use-case that one of its chained handlers requires one authority and another chained handler requires another authority. In my opinion, it doesn't make sense to set authorities on the ChainAuthHandlerImpl itself, but only on the individual auth handlers that are part of the chain. 

### Do you have a reproducer?

* https://github.com/cristic83/VertxBug

pmlopes:
Required for OpenAPI and fixes: #183

### DIFF
--- a/vertx-auth-common/src/main/java/io/vertx/ext/auth/ChainAuth.java
+++ b/vertx-auth-common/src/main/java/io/vertx/ext/auth/ChainAuth.java
@@ -29,12 +29,31 @@ public interface ChainAuth extends AuthProvider {
   /**
    * Create a Chainable Auth Provider auth provider
    *
+   * @deprecated  see {{@link #any()}}
    * @return the auth provider
    */
+  @Deprecated
   static ChainAuth create() {
-    return new ChainAuthImpl();
+    return any();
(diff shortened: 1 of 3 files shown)


pmlopes:
Fixed by d2dc036494d8bdd6921bc184721406b0f653f3b3

PR #292, opened 2019-09-04
Add basicauth for revoke call (vert.x4)

zyclonite:
same as https://github.com/vert-x3/vertx-auth/pull/291 but in master


pmlopes:
:+1:

PR #295, opened 2019-09-12
Issue/cleanup deprecations

pmlopes:


pmlopes:
According to the project governance deprecations older than 3 releases can be pruned.

PR #303, opened 2019-10-03
Mark providers and user implementation with a FQCN

pmlopes:
In a multi tenant scenario (where multiple providers are active in the same application, when user objects are deserialized over the wire there's no way to relate the source user to the original provider.

This PR will address this by adding the necessary metadata to the user object and provider so and the session level they can be matched.


pmlopes:
This PR is not applicable with the current API.

PR #304, opened 2019-10-04
Webauthn initial import

pmlopes:
This is the initial import for the webauthn/fido2 support for vertx auth.

This code already works, but it needs a review and work on the docs.


pmlopes:
Get assertion should not require username but use the rawid value. See MDN docs.

PR #305, opened 2019-10-05
fix: jwt audience unit test

lusoalex:
Hello,

The unit test was previously succeeding because the authProvider was not checking the audience.
_The audience jsonArray attribute is not used..._

To solve this, I just have done it the same way testBadAudience unit test.

Regards,
Alexandre.


pmlopes (approved):
:+1:

PR #306, opened 2019-10-05
Manage scope claim

lusoalex:
Hello,
@pmlopes I would like to have your opinion regarding scopes & JWT.

Audience ("aud") claim aim is to ensure/validate that the JWT has been created for the RS (Resource Server):
 - The Authorization Server (Okta, Auth0, Keycloak, ...) when included into and id_token.
 - The API when used as an access_token.

The audience is great when a JWT is intended to be used by several RS, we can keep control of how it is used. _Let's say we have several APIs (A,B,C,D,E...) and the JWT is intended to be used only for the APIs A,C,E, then we do expect that API B & D will reject the JWT._

---

Scopes ("scope") claim aim is to protect a resource exposed through an API endpoint.  
_For example, the "body" scope is used to protect the user body information behind /api/health/..._

I exchanged with many security people/company, the use of audience & scopes are strongly recommended. I am using Vertx and the JWTAuth handler to protect my resources. Could you have a look to this pull request, and if not convenient, could you give me some insight to do it in a proper way?

_FYI, I am also using the Vertx Open Api Contract (OpenAPI3Router), I am expecting to deep dive into this later for the same reasons._

Thanks a lot for your help,
Regards,
Alexandre.

### DIFF
--- a/vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWTAuthProviderImpl.java
+++ b/vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWTAuthProviderImpl.java
@@ -42,8 +42,6 @@ import java.security.NoSuchAlgorithmException;
 import java.security.cert.CertificateException;
 import java.util.Collections;
 import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 /**
  * @author Paulo Lopes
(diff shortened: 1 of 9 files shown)


pmlopes (commented):
I think the validation should go on the parent project vertx-jwt. In this case it would apply both to auth-jwt and auth-oauth2

pmlopes:
Regardless of the comment above I agree with this pr, thanks :+1:

lusoalex:
Thanks for your answer, I'm looking at this and come back to you once done.

lusoalex:
Hello @pmlopes ,

I move the scope validation into the JWT class from vertx-jwt.
Check is done for JWTAuthProvider & Oauth2AuthProvider, according to the JWTOptions settings.
I have added the unit test for both providers.

FYI, as the scopes may be available into the token introspection json payload, I also managed this use case.

With the current code, here is the way I do protect my Resource Server/API:
```java
JWTAuthOptions config = new JWTAuthOptions()
    .addPubSecKey(myApiConfig.getIssuerPublicKey());

config.getJWTOptions()
    .setIssuer(myApiConfig.getIssuer())
    .addAudience(myApiConfig.getAudience())
    .setScopes(myApiConfig.getScopes());

return JWTAuth.create(vertx, config);
``` 
This way, I can ensure the token: 
 - come from the right issuer (issuer)
 - has been generated to reach my API (audience)
 - **the user has granted access to his resources (scopes)**

I did not squash my commit to let you see the diff, as GitHub can do it when merging.
If you want me to squash it anyway, or have any feedbacks, please let me know.

Regards,
Alexandre.

pmlopes:
Hi @lusoalex have you signed the Eclipse ECA https://www.eclipse.org/legal/ECA.php ?

If not, please do, (it's just a simple online form) that will grant us the use of your code. Then let me know and I'll merge this PR! :+1:

lusoalex:
Hi @pmlopes, I signed the Eclipse ECA 👍

pmlopes:
Thanks!

PR #311, opened 2019-10-09
Authorization

stephanebastian:
Hello @pmlopes , 
As discussed on Gitter and Google group, here is a PR to split Authentication and Authorization.
This PR adds the interface Authorization along with implementations. It also adds the new method `Set<Authorization> User.authorizations()`

When this PR is incorporated, I will then move forward and update all Vertx-Auth projects to use the new stuff (maybe leverage the class UserImpl everywhere instead of providing their own implementation each time)

When done I will then provide another PR to update the AuthHandler to account for the new Authorization mechanism

Could you please review ? 
Thanks


vertx-auth-common/src/main/java/io/vertx/ext/auth/AndAuthorizationConverter.java
```diff
@@ -0,0 +1,42 @@
+package io.vertx.ext.auth;
```
pmlopes: this should be in a `impl` package I think, as it's some internal converter code not really public api.

vertx-auth-common/src/main/java/io/vertx/ext/auth/Authorization.java
```diff
@@ -0,0 +1,52 @@
...
+	 * this method verifies whether or not the authorization implies the specified authorization. 
+	 * Note that implies doesn't always mean an exact match. For instance, in the case of a {@link WildcardPermissionBasedAuthorization}, 
+	 * this method may return true even if the permissions are different 
+	 * => WildcardPermissionBasedAuthorization.create('*').implies(WildcardPermissionBasedAuthorization.create('anypermission')) would return true 
+	 * 
+	 * @param authorization
+	 * @return
+	 */
+	@GenIgnore
+	boolean implies(Authorization authorization);
```
pmlopes: personally i find this method name confusing, i guess other alternatives such:

* attest
* verify

would be better, but it's just IMO

vertx-auth-common/src/main/java/io/vertx/ext/auth/AuthorizationContext.java
```diff
@@ -0,0 +1,26 @@
...
+import io.vertx.core.http.HttpServerRequest;
+
+/**
+ * The AuthorizationContext contains properties that can be used to match an Authorization
+ * 
+ * @author <a href="mail://stephane.bastian.dev@gmail.com">Stephane Bastian</a>
+ *
+ */
+@VertxGen
+public interface AuthorizationContext {
```
pmlopes, line 26: I've some concerns this this interface. It assumes that authZ is http based, perhaps that should be part of vertx-web. In vertx-auth there should not be such assumption.

vertx-auth-common/src/main/java/io/vertx/ext/auth/AuthorizationConverter.java
```diff
@@ -0,0 +1,58 @@
+package io.vertx.ext.auth;
```
pmlopes: move to `impl`

vertx-auth-common/src/main/java/io/vertx/ext/auth/NotAuthorizationConverter.java
```diff
@@ -0,0 +1,31 @@
+package io.vertx.ext.auth;
```
pmlopes: move to `impl`

vertx-auth-common/src/main/java/io/vertx/ext/auth/OrAuthorizationConverter.java
```diff
@@ -0,0 +1,42 @@
+package io.vertx.ext.auth;
```
pmlopes: move to `impl`

vertx-auth-common/src/main/java/io/vertx/ext/auth/PermissionBasedAuthorizationConverter.java
```diff
@@ -0,0 +1,39 @@
+package io.vertx.ext.auth;
```
pmlopes: move to `impl`

vertx-auth-common/src/main/java/io/vertx/ext/auth/RoleBasedAuthorizationConverter.java
```diff
@@ -0,0 +1,40 @@
+package io.vertx.ext.auth;
```
pmlopes: move to `impl`

vertx-auth-common/src/main/java/io/vertx/ext/auth/WildcardPermissionBasedAuthorizationConverter.java
```diff
@@ -0,0 +1,39 @@
+package io.vertx.ext.auth;
```
pmlopes: move to `impl`

vertx-auth-common/src/test/java/io/vertx/ext/auth/ChainAuthTest.java
```diff
@@ -1,11 +1,10 @@
 package io.vertx.ext.auth;
```
pmlopes, line 1: I guess this whole file should be deprecated as `and`/`or` operations are handled by the `Authorization` interfaces right?

vertx-auth-htdigest/src/main/java/io/vertx/ext/auth/htdigest/impl/HtdigestUser.java
```diff
@@ -56,4 +60,9 @@ public JsonObject principal() {
   @Override
   public void setAuthProvider(AuthProvider authProvider) {
   }
+
+  @Override
```
pmlopes: perhaps create a default method in the interface which avoids having to patch all the sub projects?

pmlopes (commented):
The indentation is not right, you could use the `.editorconfig` present in the root of the project to fix it.
I also think the `AuthorizationContext` should not be here as it assumes http server requests.
Also how would this integrate with: https://github.com/vert-x3/vertx-auth/blob/webauthn-wip/vertx-auth-common/src/main/java/io/vertx/ext/auth/AuthStore.java which is to be added with `webauthn` as we don't want to couple it to a specific backend?

vertx-auth-common/src/test/java/io/vertx/ext/auth/ChainAuthTest.java
stephanebastian: Yes it should be deprecated

vertx-auth-common/src/main/java/io/vertx/ext/auth/AndAuthorizationConverter.java
stephanebastian: Yep, I will move all converters to 'impl'

vertx-auth-common/src/main/java/io/vertx/ext/auth/Authorization.java
stephanebastian: yes, verify seems better

vertx-auth-htdigest/src/main/java/io/vertx/ext/auth/htdigest/impl/HtdigestUser.java
stephanebastian: yep, will do

stephanebastian:
Paulo, 
Thanks for the feedback.
I believe this PR now contains fixes for all your comments
Concerning the AuthStore, I believe everything should be fine if the AuthStore is restricted to reading/storing Authentication info. Therefore I would thing the method getRoles/getPermissions are redundant with the AuthorisationProvider and thus should be removed from AuthStore. Don't you think ?

vertx-auth-common/src/main/java/io/vertx/ext/auth/Authorization.java
```diff
@@ -0,0 +1,58 @@
...
+public interface Authorization {
+
+  /**
+   * this methods verifies whether or not the authorization match the specified
+   * context.
+   * 
+   * @param context
+   * @return
+   */
+  @GenIgnore
```
pmlopes: This is a critical interface to the API, but this means it can't be used by the polyglot languages? I am guessing this would not really need to be ignored as the parameter is also annotated with `@VertxGen`

vertx-auth-common/src/main/java/io/vertx/ext/auth/Authorization.java
```diff
@@ -0,0 +1,58 @@
...
+   * </br>Note that it doesn't always mean an exact match. For instance,
+   * in the case of a {@link WildcardPermissionBasedAuthorization}, this method
+   * may return true even if the permissions are different
+   * </br>WildcardPermissionBasedAuthorization.create('*').implies(WildcardPermissionBasedAuthorization.create('anypermission')) 
+   * would return true
+   * 
+   * @param authorization
+   * @return
+   */
+  @GenIgnore
```
pmlopes: Same as above

vertx-auth-common/src/main/java/io/vertx/ext/auth/Authorization.java
stephanebastian: Yep, your are definitely right. I need to remove the @GenIgnore annotation for sure. One question though, wouldn't it be even better to make Authorization a DataObject ? If yes, if I make it a DataObject, does it need to be annotated with VertxGen to be available to all languages?

pmlopes (commented):
I think we're getting to a good shape, I'm commented on `@GenIgnore` I think they should probably be removed and it should just work as the parameters are all `@VertxGen` otherwise we can't use the API from polyglot side.

Also we need to have a copyright header in all files. Since this is a large contribution, there are a couple of extra requirements before things get merged.

1. You and all other contributors must sign the Eclipse Contributor Agreement: https://www.eclipse.org/legal/ECA.php
2. Add a copyright header to all files and list the copyright owner. To comply to the requirements please use "Copyright contributors to the Eclipse Foundation"
3. You will need to sign-off the commits with the email address you used in the ECA form.

vertx-auth-common/src/main/java/io/vertx/ext/auth/Authorization.java
pmlopes: `@DataObject` is also valid polyglot

stephanebastian:
I removed GenIgnore and added copyright headers expected for the file WildcardExpression which is a file I adapted from Apache Shiro. So I'm not the owner or initial creator of the file.
I added the original apache header. Is this ok ?

stephanebastian:
concerning the ECA, I have checked yesterday and everything should be fine.

francoisprunier:
Hi @pmlopes 

Stéphane is off at the moment. In your opinion, is this PR ready to be merged ? I'd like to get started on the next PR if everything is fine for you.

Thanks,
François

PR #321, opened 2019-11-13
Proposal to enhance User

pmlopes:
This is a proposal to enhance the User object.

* add `expired()` which shall be used by providers that issue time based tokens
* add `httpAuthorization()` which returns a HTTP Authorization header formatted for the user

The later is useful to allow the `web-client` project to just use `User` objects when creating HTTP requests. for example:

```java
User me = ... // user obtained from a JWT / OAuth2 request / etc...

client
  .get("somehost", "somepath")
  .httpAuthorization(me);
```

### DIFF
--- a/vertx-auth-common/src/main/java/io/vertx/ext/auth/User.java
+++ b/vertx-auth-common/src/main/java/io/vertx/ext/auth/User.java
@@ -16,6 +16,7 @@
 
 package io.vertx.ext.auth;
 
+import java.util.Base64;
 import java.util.Collections;
 import java.util.Set;
 
@@ -25,6 +26,7 @@ import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Promise;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.auth.impl.UserImpl;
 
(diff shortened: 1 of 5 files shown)


vertx-auth-common/src/main/java/io/vertx/ext/auth/User.java
```diff
@@ -44,13 +46,22 @@ static User create(JsonObject principal) {
...
    * @return
    */
   default Set<Authorization> authorizations() {
 	return Collections.emptySet();
   }
 
+
+  /**
+   * Assert that this user object is expired.
+   * @return true if expired
```
vietj: {@code true}

vietj: why assert ? there is no assertion / test done, it simply returns a value indicating whether the user is expired

vertx-auth-common/src/main/java/io/vertx/ext/auth/User.java
```diff
@@ -108,4 +119,23 @@ static User create(JsonObject principal) {
...
+    if (principal.containsKey("username") && principal.containsKey("password")) {
+      Buffer buff = Buffer.buffer()
+        .appendString(principal.getString("username"))
+        .appendString(":")
+        .appendString(principal.getString("password"));
+
+      return new String(Base64.getEncoder().encode(buff.getBytes()));
+    }
+
+    throw new UnsupportedOperationException("This User object doesn't support HTTP Authorization");
```
vietj: why not returning null instead ? throwing UnsupportedOperationException is quite severe here

pmlopes:
Replaced by #341

PR #322, opened 2019-11-14
JDBCAuth : Authentication and Authorization are now handled by seperate interfaces

stephanebastian:
@pmlopes Could you please review. Thx


vertx-auth-jdbc/src/main/generated/io/vertx/ext/auth/jdbc/JDBCAuthOptionsConverter.java
```diff
@@ -1,82 +0,0 @@
-package io.vertx.ext.auth.jdbc;
```
pmlopes: We can't delete this "yet", first we need to fix the code on `vertx-shell` that needs it.

vertx-auth-jdbc/src/main/java/examples/AuthJDBCExamples.java
```diff
@@ -35,7 +35,7 @@ public void example5(Vertx vertx, JsonObject jdbcClientConfig) {
 
     JDBCClient jdbcClient = JDBCClient.createShared(vertx, jdbcClientConfig);
 
-    JDBCAuth authProvider = JDBCAuth.create(vertx, jdbcClient);
+    JDBCAuthenticationProvider authenticationProvider = JDBCAuthenticationProvider.create(vertx, jdbcClient);
```
pmlopes: As we're dealing with new code, should be just drop the `Provider` suffix on all new classes? e.g.:

* `JDBCAuthentication` ?
* `JDBCAuthorization` ?

Or just go with the "regular" short name, e.g.:

* `JDBCAuthN` ?
* `JDBCAuthZ` ?

vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/JDBCAuthOptions.java
```diff
@@ -1,192 +0,0 @@
-/*
```
pmlopes: Do not remove yet... need to fix "shell" first...

Perhaps the best would be to deprecate all the setters/getters on `JDBCAuth` and use this as config instead.

vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/JDBCAuthenticationProvider.java
```diff
@@ -35,36 +35,16 @@
...
-  /**
-   * The default query to retrieve all permissions for the role
-   */
-  String DEFAULT_PERMISSIONS_QUERY = "SELECT PERM FROM ROLES_PERMS RP, USER_ROLES UR WHERE UR.USERNAME = ? AND UR.ROLE = RP.ROLE";
-
-  /**
-   * The default role prefix
-   */
-  String DEFAULT_ROLE_PREFIX = "role:";
+public interface JDBCAuthenticationProvider extends AuthProvider {
```
pmlopes: Shorter name maybe? `JDBCAuthN/JDBCAuthentication`

vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/JDBCAuthenticationProvider.java
```diff
@@ -35,36 +35,16 @@
...
 
   /**
    * Create a JDBC auth provider implementation
    *
    * @param client  the JDBC client instance
    * @return  the auth provider
    */
-  static JDBCAuth create(Vertx vertx, JDBCClient client) {
-    return new JDBCAuthImpl(vertx, client);
+  static JDBCAuthenticationProvider create(Vertx vertx, JDBCClient client) {
```
pmlopes: Maybe use the options object directly so we can avoid all the setters/getters?

vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/JDBCAuthorizationProvider.java
```diff
@@ -0,0 +1,59 @@
...
+import io.vertx.ext.auth.jdbc.impl.JDBCAuthorizationProviderImpl;
+import io.vertx.ext.jdbc.JDBCClient;
+
+/**
+ * Factory interface for creating {@link io.vertx.ext.auth.AuthorizationProvider} instances that use the Vert.x JDBC client.
+ *
+ * @author <a href="mail://stephane.bastian.dev@gmail.com">Stephane Bastian</a>
+ */
+@VertxGen
+public interface JDBCAuthorizationProvider extends AuthorizationProvider {
```
pmlopes: Shorter name? WDYT?

vertx-auth-jdbc/src/main/generated/io/vertx/ext/auth/jdbc/JDBCAuthOptionsConverter.java
stephanebastian: sure, I'll add it back

vertx-auth-jdbc/src/main/java/examples/AuthJDBCExamples.java
stephanebastian: +1: I've got a slight preference for JDBCAuthentication, JDBCAuthorization but I'm fine with AuthN, AuthZ as well

pmlopes: Let's go with the explicit then: `JDBCAuthorization / JDBCAuthentication`

vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/JDBCAuthOptions.java
stephanebastian: ok, I'll deprecate getters/setters.
Note that we probably need to support JDBCAuthenticationOptions, JDBCAuthorizationOptions and JDBCAuthOptions that implements both, WDYT ?

vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/JDBCAuthenticationProvider.java
stephanebastian: sure, same as above

vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/JDBCAuthenticationProvider.java
stephanebastian: +1: JDBCAuthentication

vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/JDBCAuthorizationProvider.java
stephanebastian: you are right, JDBCAuthorization is way better. lets got for it

vertx-auth-shiro/src/main/java/org/apache/shiro/realm/GetAuthorizationsHack.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.shiro.authz.Permission;
+import org.apache.shiro.mgt.AuthorizingSecurityManager;
+import org.apache.shiro.mgt.SecurityManager;
+import org.apache.shiro.subject.Subject;
+
+import io.vertx.ext.auth.Authorization;
+import io.vertx.ext.auth.RoleBasedAuthorization;
+import io.vertx.ext.auth.WildcardPermissionBasedAuthorization;
+
+public class GetAuthorizationsHack {
```
pmlopes, line 19: 😅 perhaps that's what it looks like but we should have a nicer name 🤷‍♂️

vertx-auth-shiro/src/main/java/org/apache/shiro/realm/GetAuthorizationsHack.java
```diff
@@ -0,0 +1,63 @@
+package org.apache.shiro.realm;
```
pmlopes, line 1: This looks like code only used internally, if a vertx dev isn't using this directly it should be in the`impl` package

vertx-auth-shiro/src/main/java/org/apache/shiro/realm/GetAuthorizationsHack.java
stephanebastian: Yep, do you have something in mind ? :)

vertx-auth-shiro/src/main/java/org/apache/shiro/realm/GetAuthorizationsHack.java
stephanebastian: The reason it is in the shiro package, and has to stay there, is that I had to use the protected method [getAuthorizationInfo(PrincipalCollection)](https://shiro.apache.org/static/1.4.1/apidocs/org/apache/shiro/realm/AuthorizingRealm.html#getAuthorizationInfo-org.apache.shiro.subject.PrincipalCollection-) from the class AuthorizingRealm... 
That's  the only solution I could find to get all authorizations at once from Shiro. Kind of hackish but get the job done..

pmlopes: humm, this will cause issues with java modules and java 11 will complain about split packages...

stephanebastian: yes, you're right.. Couldn't find a better solution yet. I'll investigate to see if there is a better way

pmlopes:
@stephanebastian can you rebase this PR on the current master so we can merge?

stephanebastian:
Paulo, I rebased it with master along with others improvements that I made recently. Please note that at this stage, the class JDBCAuth is gone so it's a breaking change. The reason was first to focus on the JDBC split. As you'll see, the JDBCClient, JDBCHashStrategy, JDBCAuthenticationOptions are now required to instantiate a JDBCAuthtenticationProvider. This is more verbose but more explicit. As a result I could remove all methods related to hashing and such from JDBCAuthentication. 
Let me know if you want me to add back the class JDBCAuth for backward compat. 
Also, to be fully complete, I need to update the project vertx-auth-oauth2 and update the user. Although this is a task that can be done later-on I guess

stephanebastian:
One more comments. Note that this PR will break vertx-shell unless you also merge https://github.com/vert-x3/vertx-shell/pull/42

pmlopes:
@stephanebastian I think we should keep the `JDBCAuth` class for now to reduce the breakage but it's fine to mark it all as `@Deprecated`

stephanebastian:
ok, let me add it back then. I already had it almost working in a commit somewhere. Let me do archeology ;), test it and we should be good

stephanebastian:
@pmlopes Here it is. I already had the code in one of my branch. Let me know if it's ok now.

pmlopes:
@stephanebastian thanks for your great work! :+1:

PR #327, opened 2019-11-22
new auth provider that read users, roles and permisions from a text file

stephanebastian:
@pmlopes Could you please review


pmlopes:
For consistency should be call it `vertx-auth-properties` ? The rationale behind it is that we also support `.htdigest`, `.htpasswd` which are also files.

stephanebastian:
sure, I updated the PR to rename the project to vertx-auth-property. I also renamed the AuthProvider to  PropertyFileAuthentication.

Issue #325, opened 2019-11-22
 Add a new provider to authenticate via LDAP - currently done thru Shiro


stephanebastian:

pmlopes:
@stephanebastian, a second task to complete this would be to `deprecate` the shiro provider (just add the annotation everywhere) and update the documentation stating to use the new properties/ldap providers instead.

stephanebastian:
sure. I opened the issue #333 for it and will do a PR asap

PR #328, opened 2019-11-25
first stab at implementation the LDAP AuthProvider

stephanebastian:
implemented #325
This implementation does authentication via LDAP. 
Note that it doesn't do authorizatiopn via LDAP. However, it's not a regression since the current LDAP provider (based on Shiro) doesn't seem to support LDAP authorization either.
@pmlopes Could you please review ?

### DIFF
--- a/vertx-auth-ldap/src/main/asciidoc/index.adoc
+++ b/vertx-auth-ldap/src/main/asciidoc/index.adoc
@@ -1,0 +1,32 @@
+== The property file auth provider
+
+This auth provider implementation reads users, roles and permissions from a property file that is compatible 
+with the format used by http://shiro.apache.org/static/1.3.2/apidocs/org/apache/shiro/realm/text/PropertiesRealm.html[Apache Shiro Properties Realm].
+
+Note that there is no dependency on Apache Shiro as this is a new implementation that is simply compatible with the Apache Shiro format
+
+The properties file should have the following structure:
+
+Each line should either contain the username, password and roles for a user or the permissions in a role.
+
+For a user line it should be of the form:
+
+ user.{username}={password},{roleName1},{roleName2},...,{roleNameN}
+
+For a role line it should be of the form:
+
+ role.{roleName}={permissionName1},{permissionName2},...,{permissionNameN}
+
+Here's an example:
+----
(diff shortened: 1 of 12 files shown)


vertx-auth-ldap/src/main/asciidoc/index.adoc
```diff
@@ -0,0 +1,32 @@
+== The property file auth provider
```
pmlopes: I think this is a copy paste error, the doc is for the properties file not ldap.

vertx-auth-ldap/src/main/java/io/vertx/ext/auth/ldap/LdapAuthentication.java
```diff
@@ -0,0 +1,41 @@
...
+public interface LdapAuthentication extends AuthProvider {
+
+  /**
+   * Create a LDAP authentication provider
+   * 
+   * @param vertx  the Vert.x instance
+   * @param options  the ldap options
+   * @return  the authentication provider
+   */
+  @GenIgnore
```
pmlopes: `options` is a data object this should not be needed.

pmlopes:
This is great and we can merge once the 2 minor points have been addressed :+1:

vertx-auth-ldap/src/main/java/io/vertx/ext/auth/ldap/LdapAuthentication.java
stephanebastian: What do you mean ? not sure I understand. 
Right now, the options is a mandatory parameter to configure the LDAP provider

vertx-auth-ldap/src/main/asciidoc/index.adoc
stephanebastian: yep definitely a cut/paste ;) I modified its content, let me know if it's ok

vertx-auth-ldap/src/main/java/io/vertx/ext/auth/ldap/LdapAuthentication.java
stephanebastian: oh got it, didn't see the GenIgnore.... Let me fix it right away

PR #332, opened 2019-11-27
User.authorizations() doesn't currently keep track of the AuthorizationProvider that contributed authorizations

stephanebastian:
@pmlopes Could you please review ?

### DIFF
--- a/vertx-auth-common/src/main/java/io/vertx/ext/auth/Authorizations.java
+++ b/vertx-auth-common/src/main/java/io/vertx/ext/auth/Authorizations.java
@@ -1,3 +1,15 @@
+/********************************************************************************
+ * Copyright (c) 2019 Stephane Bastian
+ *
+ * This program and the accompanying materials are made available under the 2
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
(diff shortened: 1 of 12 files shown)


vertx-auth-common/src/main/java/io/vertx/ext/auth/Authorizations.java
```diff
@@ -0,0 +1,24 @@
+package io.vertx.ext.auth;
```
pmlopes, line 13: header missing

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/AuthorizationsImpl.java
```diff
@@ -0,0 +1,90 @@
+package io.vertx.ext.auth.impl;
```
pmlopes, line 13: header missing

pmlopes (commented):
Just missing the copyright headers, other than that I'll just merge it ;-)

vertx-auth-common/src/main/java/io/vertx/ext/auth/Authorizations.java
stephanebastian: oups... fixed ;)

pmlopes (approved):

PR #339, opened 2019-11-29
Implement MP JWT 1.1

pmlopes:


vertx-auth-common/src/main/java/io/vertx/ext/auth/authorization/AuthorizationProvider.java
```diff
@@ -28,31 +29,31 @@
...
+//      public String getId() {
+//        return id;
+//      }
+//
+//      @Override
+//      public void getAuthorizations(User user, Handler<AsyncResult<Set<Authorization>>> handler) {
+//        handler.handle(Future.succeededFuture(new HashSet<>(_authorizations)));
+//      }
+//    };
+//  }
```
stephanebastian: Does it make sens to add the previous content in a separate class ? Something like StaticAuthorizationProvider maybe ? This is useful for tests but it might be useful for end-users as well. WDYT ?

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/UserImpl.java
```diff
@@ -45,7 +49,12 @@ public UserImpl() {
...
 
   public UserImpl(JsonObject principal) {
+    this(principal, new JsonObject());
+  }
+
+  public UserImpl(JsonObject principal, JsonObject attributes) {
     this.principal = Objects.requireNonNull(principal);
+    this.attributes = attributes;
     this.authorizations = new AuthorizationsImpl();
   }
```
stephanebastian, line 59: note a big deal but i'm wondering if we should really provide this constructor. To keep thing simple, we could only support the constructor that takes the principal parameters and add attributes by calling attributes.put(x, y)

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/UserImpl.java
```diff
@@ -54,9 +63,14 @@ public Authorizations authorizations() {
     return authorizations;
   }
 
+  @Override
+  public JsonObject attributes() {
+    return attributes;
+  }
+
   @Override
   public User clearCache() {
```
stephanebastian: the method clearCache could probably be deprecated? 
In this case we need to add the method Authorizations.delete() and cut/paste this content

stephanebastian (commented):
LGTM, cool for adding the extra attributes parameter, it's a nice addition and make thing simpler/cleaner

PR #340, opened 2019-12-02
Add User attributes

pmlopes:
Add the `attributes` property to the user interface as originally drafted on the proposal.

The attributes are a good idea that initially I did not grasp and were excluded on the pull request that followed the API enhancement.

This extra field can be useful for adding extra metadata to the user principal. For example:

* when was this user object created? (so we can perform time based expiration validation)
* where can we find `xyz` on the principal? (for token based authz different providers can hold the value to assert on different locations)
* ...


pmlopes:
@stephanebastian can you have a quick look that i did not miss anything.

pmlopes:
Fixed by #341

Issue #312, opened 2019-10-10
The wrong SQL script to create role/permissions table in jdbc-auth


Gogomoe:
I noticed that the primary key of `roles_perms` is `(role)` but not `(role, perm)` in [sql file](https://github.com/vert-x3/vertx-auth/blob/4.0.0-milestone3/vertx-auth-jdbc/src/main/resources/jdbc-auth-ddl.sql) and [docs](https://vertx.io/docs/vertx-auth-jdbc/java/#_jdbc_auth_provider_implementation).
``` sql
CREATE TABLE `roles_perms` (
  `role` VARCHAR(255) NOT NULL,
  `perm` VARCHAR(255) NOT NULL
);
ALTER TABLE roles_perms ADD CONSTRAINT `pk_roles_perms` PRIMARY KEY (role);
```

A role can only correspond to one permission according the table structure. That would make it impossible to implement permission-role model.

Is that a mistake?

pmlopes:
@stephanebastian could you verify this?

francoisprunier:
Hi @pmlopes,

I'm going to look into it.

francoisprunier:
@pmlopes I've created this PR: https://github.com/vert-x3/vertx-auth/pull/348

PR #348, opened 2019-12-12
issue #312 - fixed SQL DDL so the tuple (role,perm) is a unique key

francoisprunier:
Fixed DDL for the role_perms table, added primary keys in tests, small cleanup of comments

I had to remove a foreign key constraint on role as it not valid SQL to reference a non unique column and role is no longer unique due to the change I made (which might explain where the initial issue comes from!). For information, MySQL allows those kind of foreign key as an sql extension, but other RDMS do not.


pmlopes:
Great there's one more place to update, the asciidoc index.adoc also contains the same script if I'm not mistaken...  and the new module sql auth does have the same issue as it's a future replacement for the jdbc auth... once we have more asynchronous drivers

francoisprunier:
@pmlopes I've made the same changes to the documentation and in the new sql module.

pmlopes:
Thanks! :+1:

PR #352, opened 2019-12-15
minor comments:

stephanebastian:
Paulo, I've made a couple of small modifications mainly to clarify things a bit. 
Mainly the term 'authority' with is now an Authorization 
And also the term 'cache' that is now simply the list of authorizations

Please feel free to cherry-pick some part as you see fit.


pmlopes:
:+1:

Issue #276, opened 2019-03-22
Why Mongo Auth not to separate User-Role collection and Role-Permission collection?


UkonnRa:
### Version

* vert.x core: 3.6.3
* vert.x auth: 3.6.3

### Context

In Mongo-Auth, you can only set *ONE* collection which should contain user, role and permission field together. And in your test examples, the {user, role} and {role, premission} are combine together. I wonder is this an ideal design? Why can't I separate {user, role} and {role, premission} to two separate collections, just like JDBC's design? JDBC Auth can set `setPermissionsQuery` and `setRolesQuery`, while Mongo have not such settings...

saumiktewari:
@CasterKKK 
I too have the same problem. In fact, [here](https://vertx.io/docs/vertx-auth-mongo/kotlin/) their documentation says

"Although Vert.x auth itself does not mandate any specific model of permissions (they are just opaque strings), this implementation assumes a familiar user/role/permission model, where a user can have zero or more roles and a role can have zero or more permissions."

But their implementation puts {user, roles and permissions} in the same collection tightly coupling user with permission. This way roles become useless.

UkonnRa:
@saumiktewari Finally, I just implement `AbstractUser` and `AuthProvider` by myself, it's not difficult.

pmlopes:
Fixed by #353

PR #353, opened 2019-12-16
Mongo split authn authz

francoisprunier:
Should be similar to the JDBC auth split. I also updated the docs + examples.


francoisprunier:
Note that this PR should also resolve #276

pmlopes:
Great :+1: thanks!

PR #367, opened 2020-02-26
vertx-auth-jdbc, add exceptions to separate invalid auth info error from auth failure

xnnyygn:
Motivation:

When I learn vertx by https://vertx.io/docs/guide-for-java-devs/ ,
I find it is difficult to know the reason why user authorization was failed using vertx-auth-jdbc. Sometimes I need to response with 401 HTTP status code for JSON API request to indicate the username password pair is incorrect, and response with 500 if the underlying database is down. The only way I can do now is to check the error message, which I think is not a good way since the error message in the library may be changed in new versions.

What I did is adding two exceptions for different cases so that the client code can check the exception type to know what has happened.

Conformance:


pmlopes:
@xnnyygn  I think this PR should be handled in a different way. Instead we should have a auth info validation method that would throw exceptions like `IllegalArgumentException` or `IllegalStateException` when arguments are wrong or missing, and this would be used by all providers.

Being able to use it by all providers would also mean that exceptions would be consistent across providers, so missing a field on `jdbc` or `mongo` would result on the same exception type to be thrown.

What are your thoughts on this?

xnnyygn:
@pmlopes Sounds good. A separated validation method makes the code clean and allows fail-fast before querying the database. I will try it.

xnnyygn:
Reverted previous commit and added a new default method, validate, to the AuthenticationProvider interface. Not sure whether IllegalArgumentException is necessary or not as a part of method signature in the interface. If not, I will remove it.

Please check it when you have time.

vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/impl/JDBCAuthenticationImpl.java
```diff
@@ -56,6 +56,18 @@ public JDBCAuthenticationImpl(JDBCClient client, JDBCAuthenticationOptions optio
     this.options = Objects.requireNonNull(options);
   }
 
+  @Override
+  public void validate(JsonObject authInfo) throws IllegalArgumentException {
+    String username = authInfo.getString("username");
```
Ravattailor, line 64: what are the chances to get authInfo itself as a null?

xnnyygn: I forgot why I didn't check if the authInfo is null, but I will take your advice.

pmlopes:
We can now handle validation per provider.

Issue #369, opened 2020-03-10
OAuth2AuthHandler created from OpenIDConnectAuth does not verify audience of JWT tokens


mohamnag:
### Version

Which version(s) did you encounter this bug ?

Vert.x 3.8.5 & Vert.x 4.0.0-milestone 4

### Context

While configuring an OAuth2Auth using OIDC and setting JWT options as following, the JWT options are not taken into consideration in `OAuth2AuthProviderImpl` when a JWT token is used for authentication.

```java
    return OpenIDConnectAuth
      .rxDiscover(
        vertx,
        new OAuth2ClientOptions()
          .setSite(issuer)
          .setClientID("")
          .setJWTOptions(new JWTOptions()
            .setAlgorithm("RSD256")
            .setAudience(List.of(audience))
            .setIssuer(issuer)
            .setLeeway(10)
          )
      )
      .map(OAuth2AuthHandler::create);
```

neither audience nor the issuer set here are checked for (not even against discovered config from OIDC). Looking at `io.vertx.ext.auth.oauth2.impl.OAuth2AuthProviderImpl#authenticate` makes it clear that non of these options are taken into account when JWT is parsed successfully. The OIDC spec states that audience should be checked for.

### Do you have a reproducer?

no

### Steps to reproduce

1. setup an OIDC based auth handler as described above and mount it on some route
2. get a valid JWT token from issuer for any other audience
3. use token to call that route
4. token will be accepted and no error is thrown

### Extra

not relevant

pmlopes:
Fixed by #370

PR #370, opened 2020-03-11
Validate the issuer and audience claims

pmlopes:
Motivation:

Currently Auth JWT and OAuth2 are not performing the same validations, this PR ensures that both perform the spec required validations equally.


pmlopes:
Fixes #369

PR #379, opened 2020-03-30
MongoUserUtil: pass the created document identifiers

jponge:
This brings the behavior of `MongoUserUtil` closer to the previous `insertUser` and co APIs.

This pull-request also:

- upgrades the embedded MongoDB server, and
- fixes a bug when creating users where the password ends up in the user name.

### DIFF
--- a/vertx-auth-mongo/pom.xml
+++ b/vertx-auth-mongo/pom.xml
@@ -38,7 +38,7 @@
     <dependency>
       <groupId>de.flapdoodle.embed</groupId>
       <artifactId>de.flapdoodle.embed.mongo</artifactId>
-      <version>2.0.3</version>
+      <version>2.2.0</version>
       <scope>test</scope>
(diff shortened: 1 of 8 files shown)


jponge:
The `vertx-auth-sql` module needs an upgrade to the `vertx-sql-client` latest API changes. Doing that.

jponge:
Done

pmlopes:
Maybe a next step would be get rid of flapdoodle and use test containers

jponge:
> Maybe a next step would be get rid of flapdoodle and use test containers

Absolutely, but let's defer that to another PR

jponge:
@pmlopes Do the proposed changes look good to you? (aka do you mind if I merge now?)

pmlopes (approved):

jponge:
Thanks @pmlopes!

Issue #383, opened 2020-04-15
Improve the JWT verification exception reporting


sberyozkin:
Right now we have to check an exception message content to figure out if the verification failure was due to the token expiry or not.
I guess we can try to introduce a `JwtVerificationException` and may be 2 subclasses like `JwtSignatureVerificationException` and `JwtExpiryException` or may be have a single extension and enum based property to indicate the actual problem. I can try a quick PR

sberyozkin:
Hi Paulo, will try this one as well

PR #387, opened 2020-05-08
Forced JWK refresh if no matching JWK is available

### LINKED ISSUE #382: Improve the JWK based verification code
Right now all the keys from the JWK set are iterated over in order to verify the token, and when the verification fails it is not possible in Quarkus to determine if the verification failed because there was no matching `kid` or it was something else => we can't do the on-demand JWKS load reliably.

I'd like to do a small PR with some minor updates

sberyozkin:
Fixes #382 #383

Motivation:

As explained in #382 and #383:

- We currently check for the text messages like "expired token" in order to figure out that the token verification failed because the token has expired, so `JWTException` (extending original `IllegalStateException`) is introduced with an initial subset of the `Reason` enum which can be enhanced as needed going forward.
- The JWT verification code is updated to report that the signature failed due to a missing JWK when it was detected that the token kid did not match any of the available JWKs (I haven't gone ahead enforcing that only a matching JWK can be used as I think it may break some users code, given that in some cases such as when PublicKey is added, etc, there may be no match - can be reviewed in the next phase how to make it stricter, may be with a property)
- Updated the code to do a forced refresh if no JWK with a matching kid is available. Without this forced refresh, what happens is that once the key rotation is done, every token verification will become very suboptimal:
  * first the local JWK verification will fail
  * next the remote introspection will be done

This is something we've discussed with Pedro - so if this no matching kid JWK issue is detected then we do a controlled refresh, time scoped, to avoid DOS attacks as recommended to me by @stianst and Quarkus user. This interval should be configurable (Pedro, Stian, what is the best default value ) as an option - I'll add it once we agree on it.

So this forced refresh will get the rotated keys and the remote introspection may be avoided.

We can do this forced refresh in Quarkus too but it feels it can be safely done in Vertx OAuth2, it feels it belongs here.

Paulo, @pmlopes, once you are happy with the PR idea in principle I'll then add a test emulating the key rotation situation and open PR for review. 

Thanks all

CC @pedroigor @stianst

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java
@@ -15,23 +15,29 @@
  */
 package io.vertx.ext.auth.oauth2.impl;
 
+import java.util.Collections;
+
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.impl.logging.Logger;
 import io.vertx.core.impl.logging.LoggerFactory;
+import io.vertx.core.json.DecodeException;
 import io.vertx.core.json.Json;
 import io.vertx.core.json.JsonArray;
-import io.vertx.core.json.DecodeException;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.auth.PubSecKeyOptions;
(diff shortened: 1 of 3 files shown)


sberyozkin:
@pmlopes If you don't want the forced refresh check be done in Vertx then we need a way to do it in Quarkus when the verification has failed due to a non-matching JWK without an introspection. See the reasons why above. In that case we'd need an exception be propagated when no matching JWK exception occurs.

Let me know please what is your preference, thanks

sberyozkin:
@pmlopes Hi, happy to close this one as a more robust solution is available after your merges, cheers

PR #389, opened 2020-05-12
Issues/oidc sec10 1 1 sec10.2.1

pmlopes:
Motivation:

OIDC/OAuth2 servers can expose a jwk set endpoint where keys used to validate keys are published. Currently the rotation of these keys is not considered and it is up to the user to perform this task.

However after discussion with the community the OIDC spec defines how clients should handle this, and this PR implements/test it.

Community discussion: https://groups.google.com/forum/#!topic/vertx/1O3SDsQXgQ8
OIDC Core spec: https://openid.net/specs/openid-connect-core-1_0.html#RotateEncKeys


pmlopes:
replaces #387

PR #399, opened 2020-05-25
Strongly typed auth info

stephanebastian:
@pmlopes As planned, I've implemented AuthInfo for the following providers:
- Htdigest
- Htpasswd
- JDBC
- JWT
- LDAP

While I still have to finish MongoDb (will work on it tonight), could you please do review ?


stephanebastian:
Paulo, one more thing concerning naming convention: 
I propose to rename AuthInfo to AuthenticationInfo to avoid confusion. WDYT ?

pmlopes:
@stephanebastian I think it is a good idea. Maybe a more common name would be `Subject`?

e.g.:

```java
authenticationProvider.authenticate(
new XyzSubject()
  .setUsername("paulo")
  .setPassword("secret")));
```

WDYT? Where `Xyz` is the provider prefix, e.g.: `MongoSubject`, `JWTSubject`, etc...

stephanebastian:
Don't want to be picky but I feel that terms that are commonly used in existing security libraries (such as javax.security.auth, Shiro, etc) already carry some meaning for most people. For instance to me Subject is the output of the Authentication process. Whereas Credentials is the input. So I would be more inclined to pick Credential instead of Subject ;)

This make me think that we don''t necessarily need to have a specific AuthInfo per AuthenticationProvider. We could have UsernamePasswordAuthenticationInfo (or UsernamePasswordSubject, UsernamePasswordCredential, UsernamePassword<ChooseADifferentName>, etc) used by several authentication providers. DYT ?

pmlopes:
I think it's a good idea, we can add a generic `Credential` under `vertx-auth-common` module:

`UsernamePasswordCredentials` that shall be used for most of the providers, `jdbc`, `mongo`, `ldap`, etc... and for specials ones `JWT` `OAuth2`, `webauthn`, etc.... they will have it's custom ones

stephanebastian:
I'll make the changes later tonight.

stephanebastian:
Paulo, I realized that we probably want to update the various handlers so that they call the strongly typed version of the method authenticate. We don't strictly have to do it to make it work but it would make sens. WDYT ?

PR #427, opened 2020-10-09
normalize the attributes layout and extract the userInfo specific att…

### LINKED ISSUE #405: user profiles API for Oauth2 (maybe common)?
Many of the oauth2 providers out there, have an extra API to fetch a user profile. A user profile usually contains:

* username
* name
* email
* photo

How interesting would it be to have a DataObject with this type safe structure and an extra method on the Oauth2 provider to fetch the profile?

```java
void fetchProfile(User user);
```

Perhaps we could even "standardize" on the other providers to also have this? for example, jdbc/sql/mongo could have an extra optional query to fetch this data?

The reason for the parameter is that we can use it to fetch the data directly if in the attributes/principal or using the knowledge on the provider fetch the user specific data using its unique identifier.

As it is being described here, it's a purely fetch, there is no intent to create a full CRUD api for it.

/cc @stephanebastian @jponge @francoisprunier 

The use case I can see for this is building webapps, where in many cases it is common to have a menu with the picture of the user and at least this information. Having it as a common API could simplify this use case.

pmlopes:
…ributes from the json to the user object

Related to #405 

/cc @stephanebastian @francoisprunier 

This is just a small stab at the issue. The `userInfo` API now extracts the basic attributes to the `user` instance. Perhaps we could define some common getters? or just rely on `user.attributes.getString("sub")` for example...

### DIFF
--- a/vertx-auth-common/src/main/java/io/vertx/ext/auth/User.java
+++ b/vertx-auth-common/src/main/java/io/vertx/ext/auth/User.java
@@ -40,10 +40,43 @@ import io.vertx.ext.auth.impl.UserImpl;
 @VertxGen
 public interface User {
 
+  /**
+   * Factory for user instances that are single string. The credentials will be added to the principal
+   * of this instance. As nothing can be said about the credentials no validation will be done.
+   *
+   * The kind will be used a the property key name on the principal to store the value. For example:
+   *
+   * {@code create("access_token", jwt)}
+   *
+   * Will create a principal with a property {@code "access_token"} with the jwt as value.
+   *
+   * @param kind the credential kind e.g.: {@code access_token} or {@code username}
+   * @param value the value for this token
+   * @return user instance
+   */
+  static User create(String kind, String value) {
+    return create(new JsonObject().put(kind, value));
+  }
+
+  /**
+   * Factory for user instances that are free form. The credentials will be added to the principal
+   * of this instance. As nothing can be said about the credentials no validation will be done.
+   *
+   * @param principal the free form json principal
+   * @return user instance
+   */
   static User create(JsonObject principal) {
     return create(principal, new JsonObject());
   }
 
+  /**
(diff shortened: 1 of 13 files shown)


vertx-auth-common/src/main/java/io/vertx/ext/auth/User.java
```diff
@@ -40,10 +40,43 @@
...
+   *
+   * {@code create("access_token", jwt)}
+   *
+   * Will create a principal with a property {@code "access_token"} with the jwt as value.
+   *
+   * @param kind the credential kind e.g.: {@code access_token} or {@code username}
+   * @param value the value for this token
+   * @return user instance
+   */
+  static User create(String kind, String value) {
```
stephanebastian: I am wondering if it make sens to convert kind to an enum (AccessToken, Username and FreeForm) ? In that case, we probably want to add the method User.kind() as well

stephanebastian (commented):
Other than that, it looks like a nice little improvement over the current version.

vertx-auth-common/src/main/java/io/vertx/ext/auth/User.java
pmlopes: The 2 most common cases are:

```java
create("username")
create("token")
```

I wanted to make a distinction between the 2, so I added a "kind", A second alternative would be a distict factory name:

```java
User.forName("username")
User.forToken("token)
```

Maybe this would make it even more obvious? @stephanebastian  wdyt?

pmlopes:
```java
User.fromName() / User.fromToken()
```

`for`/`from` prefix?

stephanebastian: In that case, IMHO the factory method is much better. As tfor the name I've got a slight preference for 'User.fromName() and 'User.fromToken()' but for will do as well ;)

PR #428, opened 2020-10-09
Extend credentials for client usage

pmlopes:
Motivation:

Attempt to use `Credentials` both for server side validation and for client authentication

### DIFF
--- a/vertx-auth-common/src/main/java/io/vertx/ext/auth/authentication/TokenCredentials.java
+++ b/vertx-auth-common/src/main/java/io/vertx/ext/auth/authentication/TokenCredentials.java
@@ -13,6 +13,8 @@
 package io.vertx.ext.auth.authentication;
 
 import io.vertx.codegen.annotations.DataObject;
+import io.vertx.core.Vertx;
+import io.vertx.core.http.HttpMethod;
 import io.vertx.core.json.JsonObject;
 
 /**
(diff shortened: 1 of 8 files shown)


vertx-auth-common/src/main/java/io/vertx/ext/auth/authentication/TokenCredentials.java
```diff
@@ -63,4 +63,9 @@ public JsonObject toJson() {
   public String toString() {
     return toJson().encode();
   }
+
+  @Override
+  public String toHttpHeader() {
```
pmlopes: We could have an overload with extra parameters like: method, uri here to support complex encodings like digest auth.

vertx-auth-htdigest/src/main/java/io/vertx/ext/auth/htdigest/HtdigestCredentials.java
```diff
@@ -172,4 +244,81 @@ public JsonObject toJson() {
...
+  @Override
+  public synchronized String toHttpHeader() {
+    byte[] ha1 = MD5.digest(String.join(":", username, realm, password).getBytes(StandardCharsets.UTF_8));
+    byte[] ha2 = MD5.digest(String.join(":", method, path).getBytes(StandardCharsets.UTF_8));
+
+    String cnonce;
+    String nc;
+
+    if (qop == null || "auth".equals(qop)) {
+      cnonce = random.nextString(8);
```
pmlopes: This is wrong, qop == null is valid but no nonce should be generated.

vertx-auth-htdigest/src/main/java/io/vertx/ext/auth/htdigest/HtdigestCredentials.java
```diff
@@ -13,19 +13,44 @@
...
+
+  static {
+    try {
+      MD5 = MessageDigest.getInstance("MD5");
+    } catch (NoSuchAlgorithmException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static int NC = 0;
```
pmlopes: This isn't correct, NC should be client specific

pmlopes (commented):
API isn't correct

pmlopes:
@hectorvent This PR ensures that the auth module can handle `Basic`, `Bearer` and `Digest` HTTP authentication both for client and server.

Can you please review.

pmlopes:
The implementation is not fully complete. The HTDigest module could get some update to support:

https://tools.ietf.org/html/rfc7616

Where the hardcoded `MD5` hash can be replaced by: "SHA-256", "SHA-256-sess", "SHA-512" and "SHA-512-sess"

Adding suport for this could be a good new feature to implement

pmlopes:
I've clean up all non related code changes.

Issue #441, opened 2020-12-18
UserConverter does not take into account the attributes field


thced:
Version 4.0.0

### Context

While serializing a user to/from Hazelcast via ClusteredSessionStore, I noticed that when encoding/decoding the User (UserImpl) there was information available on the User when encoding it, but that information is lost (and not even available) when decoding it. The `attributes` field is not considered at all in the UserConverter (by first, simply looking at the static fields in the class).

pmlopes:
Fixed by #442

PR #442, opened 2020-12-18
fix: UserConverter does not consider attributes

thced:
Motivation:

When encoding/decoding the User (in this case via ClusteredSessionStore), the attributes of a User is lost upon encoding, and therefore not available after decoding.


pmlopes:
:+1: Thanks!

thced:
fixes #441

PR #446, opened 2021-01-02
Change error message for expired jwt token

u6f6o:
Motivation:

The current error message _Invalid JWT token: missing required scopes._ is miss-leading as it does not inform about the expired token.


pmlopes:
Thanks!

PR #447, opened 2021-01-07
Get the scope from the user not the principal

ikstewa:
Depending on the AuthenticationProvider impl the scope may not be in the principal. Instead use the User#get method which checks multiple locations. This allows for the ScopeAuthorization to be used in conjunction with the OAuth2Auth.

Conformance:


pmlopes:
I believe all providers now store the scopes in the principal.

PR #449, opened 2021-01-11
Secret materials encoded as string assume UTF8 instead of US_ASCII

### LINKED ISSUE #448: JWT verification is broken if secret contains only non-ascii characters
### Version

4.0.0

### Context

We are using random byte arrays for our symmetric key secrets. I tried to replace our existing jwt verification, which is using [jjwt](https://github.com/jwtk/jjwt) by vertx solutions instead. Integrating these solutions, I recognised, that my jwts which are issued by another service cannot be verified. 

After some digging in the vert.x code I found out, that the key secret is transformed into an us-ascii byte array [here](https://github.com/vert-x3/vertx-auth/blob/7e94c383479a02e3e400d2ce3e755a3f028ce87a/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java#L183 ). This is bad, as every character of my secret was replaced by (byte)63. 

This is a security issue as well, as the verification can be easily broken if the secret contains only of non-ascii characters, as shown in my reproducer. 

### Do you have a reproducer?

Yes, [here](https://github.com/u6f6o/vertx-jwt-bug/blob/master/src/test/kotlin/JwtTest.kt). 

### Steps to reproduce

1. Clone the project
2. Fetch dependencies using gradle
3. Run JwtTest.kt

pmlopes:

Fixes #448 

Secret materials should assume UTF8 instead of US_ASCII

### DIFF
--- a/vertx-auth-common/src/main/generated/io/vertx/ext/auth/PubSecKeyOptionsConverter.java
+++ b/vertx-auth-common/src/main/generated/io/vertx/ext/auth/PubSecKeyOptionsConverter.java
@@ -23,7 +23,7 @@ public class PubSecKeyOptionsConverter {
           break;
         case "buffer":
           if (member.getValue() instanceof String) {
-            obj.setBuffer((String)member.getValue());
+            obj.setBuffer(io.vertx.core.buffer.Buffer.buffer(JsonUtil.BASE64_DECODER.decode((String)member.getValue())));
           }
           break;
         case "certificate":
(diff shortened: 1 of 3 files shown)


vietj:
@pmlopes LGTM however can you create a breaking change page for 4.0.1 and add this change ? we need also to check it's fine for the vertx in action examples, ping @jponge

jponge:
It shouldn't impact the book examples here.

vietj:
the breaking change is `public Buffer getBuffer` that used to return a string

vietj:
thanks @jponge

jponge:
That being said it should be tested against https://github.com/jponge/vertx-in-action/tree/master/part2-steps-challenge

jponge:
Let me test this branch

jponge:
You know what?








It worked 😉

pmlopes:
@jponge @vietj I've added it to the wiki. I checked the API, this change should not affect any user code. The getter is only used internally by the `JWT` parser/encoder. The setter, however, is used by users, that is why I kept the existing one and added the overload to `Buffer`.

The only case where this change may affect user code is when the users are implementing themselves the JWT parsing and reuse the vertx-auth-common configuration objects. I did (quickly) grep GH and did not found any uses of that method.

PR #460, opened 2021-02-19
added support for additional type definitions #459

### LINKED ISSUE #459: JWT public and private key parsing issue
Hi,

We run into an issue using JWT in our product, because loading of the public key ran into the exception. We added a workaround for it, but it might be worth fixing.

In the switch/case there is case for `PRIVATE RSA KEY` and `PUBLIC RSA KEY`, which seems not to be the default format for the pre/postfix.
The default format seems to be `RSA PRIVATE KEY` and `RSA PUBLIC KEY`.

https://github.com/vert-x3/vertx-auth/blob/ee361aeec879f2e707d372883efb1b3214aecde1/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java#L304

https://github.com/vert-x3/vertx-auth/blob/ee361aeec879f2e707d372883efb1b3214aecde1/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java#L308

tipok:
* added the support for RSA PUBLIC KEY and RSA PRIVATE KEY

Motivation:

Fixes #459

Conformance:

### DIFF
--- a/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java
+++ b/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java
@@ -302,10 +302,12 @@ public final class JWK implements Crypto {
         return USE_ENC;
       case "PUBLIC KEY":
       case "PUBLIC RSA KEY":
+      case "RSA PUBLIC KEY":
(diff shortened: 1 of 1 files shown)


pmlopes:
Thanks! :+1:

PR #474, opened 2021-04-09
Initial support for EdDSA

pmlopes:

Motivation:

Add support for EdDSA keys

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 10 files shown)


pmlopes:
@vietj I think we should keep it. We test 8 and 11 because they are the lts versions. Then 15 because it includes the EdDSA bits to test the latest signature algorithms. Btw I had to downgrade to 15 because codegen didn't want to work.

pmlopes:
Will keep 11 for the reason that the same kind of testing is used to verify the `PSxxx` algorithm family which is JDK11 and up only.

Issue #483, opened 2021-05-19
Vertx Auth JWT - Chaining providers


lusoalex:
### Describe the feature

I need to chain two JWT authentication providers, so I followed this documentation: https://vertx.io/docs/vertx-auth-common/java/#_chaining_authentication_providers

#### Step one: Create first JWTAuth

The first AuthProvider is a JwtAuth created through io.vertx.ext.auth.jwt.JWTAuth.create resulting in a `JWTAuthProviderImpl` object.

#### Step two: Create the second JWTAuth

The second one is a custom JWTAuth (named JWTClaimsVerifier), which basically, adds some extra check on claims in addition to what is already done by the previous JWTAuth (exp, nbf, etc...). 
_Ex: ensure there is no `ati` claims (some authorization servers use JWT as refresh token instead of opaque tokens, so they add an **_ati_** `access token identifier` claim referring to the "true" access_token identifier, aka `jti` that has been generated in the same time). As a result, the difference between refresh and access Json Web Token is the presence of the `ati` claim on refresh._ In my case, the custom JWTAuth will reject such refresh tokens. _There are some others claims property checks specific to our company business case, not relevant for this request._

#### Step three: Chain the authentication providers

Here I followed the documentation mentioned above by using: ChainAuth.all() and adding the two JWTAuth, resulting in a ChainAuth authentication provider.

#### Last step: Creating an Authentication Handler.

I can't use the JwtAuthHandler as it expects a JWTAuth to create a JWTAuthHandlerImpl and I have now a ChainAuth.

I wanted to create my own handler `JWTChainAuthHandlerImpl` like this
```JAVA
 JWTChainAuthHandlerImpl extends HTTPAuthorizationHandler<ChainAuth> 
```
but the HTTPAuthorizationHandler is not public.

I do not want to duplicate the code of the HTTPAuthorizationHandler, extracting token from an Authorization header is something basic, so I was wondering if : 

 - it would be possible to make HTTPAuthorizationHandler public
 - or it would be possible to add a JWTChainAuthHandler into the Vertx library?

### Contribution

For both suggestions or any other from your part, I am willing to contribute.

### How do I manage this currently...

I have created a custom chain authentication like this : 
```JAVA
public class JWTChainAuthProvider implements ChainAuth, JWTAuth {
```
[...]

PR #486, opened 2021-06-06
Keep user attributes while merge in results of a chain authentication handler

lusoalex:
# Motivation

As mentioned here #483 , I meet some issues while upgrading to the 4.1.0

With the last release, the JwtAuthHandlerImpl has setup a [postAuthentication](https://github.com/vert-x3/vertx-web/blob/master/vertx-web/src/main/java/io/vertx/ext/web/handler/impl/JWTAuthHandlerImpl.java#L117) based on a `accessToken` retrieved from the user context.

This `accessToken` is [injected into the user attributes](https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWTAuthProviderImpl.java#L154) when the JWTAuth succeed.

But in the case of chain auth, [only the user principal is merged](https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/ChainAuthImpl.java#L86), not the attributes.

As a consequence, the upgrade from 4.0.3 to 4.1.0 was failing on my side.

Please note that I have a custom Chain Handler (because currently there is no easy way to chain two JWTAuth, but reported my fix here in case there are some other issues related to missing user attributes.

Conformance:

### DIFF
--- a/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/ChainAuthImpl.java
+++ b/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/ChainAuthImpl.java
@@ -83,7 +83,7 @@ public class ChainAuthImpl implements ChainAuth {
           resultHandler.handle(res);
         } else {
           // if ALL then a success check the next one
-          iterate(idx + 1, authInfo, resultHandler, previousUser == null ? res.result() : User.create(previousUser.principal().mergeIn(res.result().principal())));
+          iterate(idx + 1, authInfo, resultHandler, merge(previousUser, res.result()));
         }
       } else {
         // try again with next provider
@@ -97,5 +97,15 @@ public class ChainAuthImpl implements ChainAuth {
         }
       }
     });
+  }
+
+  private User merge(User previousUser, User resultUser) {
+    if(previousUser==null) {
+      return resultUser;
+    }
+    return User.create(
+      previousUser.principal().mergeIn(resultUser.principal()),
+      previousUser.attributes().mergeIn(resultUser.attributes())
+    );
   }
 }
--- a/vertx-auth-common/src/test/java/io/vertx/ext/auth/ChainAuthTest.java
+++ b/vertx-auth-common/src/test/java/io/vertx/ext/auth/ChainAuthTest.java
@@ -167,17 +167,20 @@ public class ChainAuthTest {
 
     auth.add((authInfo, res) -> {
       // always OK
-      res.handle(Future.succeededFuture(createUser(new JsonObject().put("provider", 1))));
+      res.handle(Future.succeededFuture(User.create(new JsonObject().put("provider", 1), new JsonObject().put("attribute", "one"))));
     });
 
     auth.add((authInfo, res) -> {
       // always OK
-      res.handle(Future.succeededFuture(createUser(new JsonObject().put("provider", 2))));
+      res.handle(Future.succeededFuture(User.create(new JsonObject().put("provider", 2), new JsonObject().put("attribute", "two"))));
     });
 
     auth.authenticate(new JsonObject(), res -> {
       if (res.succeeded()) {
+        User result = res.result();
+        should.assertNotNull(result);
         should.assertEquals(2, res.result().principal().getInteger("provider").intValue());
+        should.assertEquals("two",res.result().attributes().getString("attribute"));
         test.complete();
       } else {
         should.fail();


lusoalex:
@pmlopes I was wondering if it would not be more convenient to have a merge function into the User returning a new instance, instead of handling this into the Chain (and maybe somewhere else?)

pmlopes:
@lusoalex yes I can imagine that this feature can also be useful for "on-behalf-of" oauth2 flows, where after an initial authn you have a user (`user1`) and with that user token, you request a second token to execute actions on-behalf-of some other service (`user2`) you may want to merge, the 2 but it can be tricky in this case as now you have 2 `access_token` in the principal which would overwrite each other.

What should be the behavior? throw some exception? (principal can't overwrite existing properties) ignore + logging?

pmlopes:
@lusoalex if you sign the eclipse ECA and do a force push the failure check should clear

http://www.eclipse.org/legal/ECA.php

lusoalex:
> @lusoalex if you sign the eclipse ECA and do a force push the failure check should clear
> 
> http://www.eclipse.org/legal/ECA.php

Lo, yes I had a look, I did already sign them, but i change the git email to the noreply github one, so ECA is not matching.
I'll solve it soon ;)

lusoalex:
> @lusoalex yes I can imagine that this feature can also be useful for "on-behalf-of" oauth2 flows, where after an initial authn you have a user (`user1`) and with that user token, you request a second token to execute actions on-behalf-of some other service (`user2`) you may want to merge, the 2 but it can be tricky in this case as now you have 2 `access_token` in the principal which would overwrite each other.
> 
> What should be the behavior? throw some exception? (principal can't overwrite existing properties) ignore + logging?

@pmlopes I am torn between failing because there are no reasons to have different access_token values (and by the way, are there other properties to check too). And just ignore because it seems obvious to me that if I merge user instance B on user instance A, B properties will overwrite A values in case of common key...

In the context of ChainAuth, I do prefer to ignore it, plus, it will avoid breaking change.

According to your feedback, I'll update this PR by moving the merge function into the User.

pmlopes:
@lusoalex I think we can settle in your initial idea of, plain merge. My concern was about:

* https://datatracker.ietf.org/doc/html/draft-ietf-oauth-token-exchange-10
* https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow

In these flows, a user impersonates a second identity, so in reality we can't/should merge, but keep them separated. Maybe a second task would be to add some stack like functionality, so a when a user impersonates a second identity, the original identity is "pushed" down the stack and when needed, we can "restore" (pop from the stack).

pmlopes:
@lusoalex sorry for the delay. I've been discussing some improvements to the `User` API and a few use cases have come up. Say for example, that there are 2 user objects:

```js
userA = {
  attributes: {
    roles: [ 'read' ]
  }
}

userB = {
  attributes: {
    roles: [ 'write' ]
  }
}
```

If we just use the JSON API `json.merge(JsonObject)` we end up with:

```js
userA.merge(userB);
// results in
{
  attributes: {
    roles: [ 'write' ]
  }
}
```

Because the JSON API only merges map structures, should we also consider `arrays` for this specific case?

For example:

```js
userA.merge(userB);
// results in
{
  attributes: {
    roles: [ 'read', 'write' ]
  }
}
```

lusoalex:
> @lusoalex sorry for the delay. I've been discussing some improvements to the `User` API and a few use cases have come up. Say for example, that there are 2 user objects:
> 
> ```js
> userA = {
>   attributes: {
>     roles: [ 'read' ]
>   }
> }
> 
> userB = {
>   attributes: {
>     roles: [ 'write' ]
>   }
> }
> ```
> 
> If we just use the JSON API `json.merge(JsonObject)` we end up with:
> 
> ```js
> userA.merge(userB);
> // results in
> {
>   attributes: {
>     roles: [ 'write' ]
>   }
> }
> ```
> 
> Because the JSON API only merges map structures, should we also consider `arrays` for this specific case?
> 
> For example:
> 
> ```js
> userA.merge(userB);
> // results in
> {
>   attributes: {
>     roles: [ 'read', 'write' ]
>   }
> }
> ```

Hi, @pmlopes I was thinking about this, by having in mind that the user is and must be the same, I do not see reasons to remove previous roles in your example. When chaining two auth providers, JWT roles payload and "local database ACL",  I would expect to keep both values. What's your opinion?

pmlopes:
@lusoalex yes, we should give a try to keep both values. This means we need to have a custom `merge()` function that works almost like the `JsonObject.mergeIn(JsonObject)` with the difference that when dealing with `arrays` instead of taking the latest, would make a copy of the source array and do a `addAll()` from the second.

As this is very specific to this use case, I think it should be implemented in this module, not on `vertx-core`.

pmlopes:
If this is too complex for now, we can start with the current implementation and improve it later if this is really necessary.

PR #488, opened 2021-06-14
feat: added otp module with hotp auth implementation

otricadziziusz:
### Motivation:

I create sub module auth-otp with implement HOTP(rfc4226) authentication. I want share with vertx community and a little bit later want support TOTP(rfc6238) authenticator.
Module support Google Authenticator with base32 key and qr codes. Could set auth attempts limit and resynchronization counter.

### DIFF
--- a/vertx-auth-otp/pom.xml
+++ b/vertx-auth-otp/pom.xml
@@ -26,13 +26,6 @@
 
   <artifactId>vertx-auth-otp</artifactId>
 
-  <repositories>
-    <repository>
-      <id>jitpack.io</id>
-      <url>https://jitpack.io</url>
-    </repository>
-  </repositories>
-
   <properties>
     <doc.skip>false</doc.skip>
   </properties>
@@ -53,11 +46,6 @@
       <groupId>commons-codec</groupId>
       <artifactId>commons-codec</artifactId>
       <version>1.15</version>
-    </dependency>
-    <dependency>
-      <groupId>com.github.otricadziziusz</groupId>
-      <artifactId>java-otp</artifactId>
-      <version>generate_strings-SNAPSHOT</version>
     </dependency>
     <dependency>
       <groupId>io.vertx</groupId>
--- a/vertx-auth-otp/pom.xml
+++ b/vertx-auth-otp/pom.xml
@@ -47,5 +47,10 @@
       <artifactId>commons-codec</artifactId>
       <version>1.15</version>
     </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-unit</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
--- a/vertx-auth-otp/pom.xml
+++ b/vertx-auth-otp/pom.xml
@@ -60,11 +60,6 @@
       <version>generate_strings-SNAPSHOT</version>
     </dependency>
     <dependency>
-      <groupId>com.github.kenglxn.qrgen</groupId>
-      <artifactId>javase</artifactId>
-      <version>2.6.0</version>
-    </dependency>
-    <dependency>
       <groupId>io.vertx</groupId>
       <artifactId>vertx-unit</artifactId>
       <scope>test</scope>
--- a/vertx-auth-otp/pom.xml
+++ b/vertx-auth-otp/pom.xml
@@ -1,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2014 Red Hat, Inc.
+  ~
+  ~  All rights reserved. This program and the accompanying materials
+  ~  are made available under the terms of the Eclipse Public License v1.0
+  ~  and Apache License v2.0 which accompanies this distribution.
+  ~
+  ~  The Eclipse Public License is available at
+  ~  http://www.eclipse.org/legal/epl-v10.html
+  ~
+  ~  The Apache License v2.0 is available at
+  ~  http://www.opensource.org/licenses/apache2.0.php
+  ~
+  ~  You may elect to redistribute this code under either of these licenses.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>vertx-auth</artifactId>
+    <groupId>io.vertx</groupId>
+    <version>4.1.1-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>vertx-auth-otp</artifactId>
+
+  <repositories>
+    <repository>
+      <id>jitpack.io</id>
+      <url>https://jitpack.io</url>
+    </repository>
+  </repositories>
+
+  <properties>
+    <doc.skip>false</doc.skip>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-auth-common</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-core</artifactId>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
+    <!-- https://mvnrepository.com/artifact/commons-codec/commons-codec -->
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+      <version>1.15</version>
+    </dependency>
+    <dependency>
+      <groupId>com.github.otricadziziusz</groupId>
+      <artifactId>java-otp</artifactId>
+      <version>generate_strings-SNAPSHOT</version>
+    </dependency>
+    <dependency>
+      <groupId>com.github.kenglxn.qrgen</groupId>
+      <artifactId>javase</artifactId>
+      <version>2.6.0</version>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-unit</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+</project>
--- a/vertx-auth-otp/pom.xml
+++ b/vertx-auth-otp/pom.xml
@@ -47,10 +47,5 @@
       <artifactId>commons-codec</artifactId>
       <version>1.15</version>
     </dependency>
-    <dependency>
-      <groupId>io.vertx</groupId>
-      <artifactId>vertx-unit</artifactId>
-      <scope>test</scope>
-    </dependency>
   </dependencies>
 </project>
--- a/vertx-auth-otp/src/main/java/examples/hotp/HotpAuthExamples.java
+++ b/vertx-auth-otp/src/main/java/examples/hotp/HotpAuthExamples.java
@@ -12,7 +12,7 @@
 package examples.hotp;
 
 import io.vertx.core.json.JsonObject;
-import io.vertx.ext.auth.impl.UserImpl;
+import io.vertx.ext.auth.User;
 import io.vertx.ext.auth.otp.hotp.HotpAuth;
 import io.vertx.ext.auth.otp.hotp.HotpAuthOptions;
 import io.vertx.ext.auth.otp.utils.OtpKey;
@@ -37,18 +37,18 @@ public class HotpAuthExamples {
     final OtpKey otpKey = new OtpKey(keyBytes, "HmacSHA1");
 
     // request hotp for user
-    JsonObject principal = new JsonObject() {{
-      put("identifier", "user1");
-      put("key", otpKey.getBase32());
-      put("counter", 0);
-    }};
-    authProvider.requestHotp(new UserImpl(principal), userAsyncResult -> {});
+    JsonObject principal = new JsonObject()
+      .put("identifier", "user1")
+      .put("key", otpKey.getBase32())
+      .put("counter", 0);
+
+    authProvider.requestHotp(User.create(principal), userAsyncResult -> {});
 
     // auth user hotp
-    JsonObject credentials = new JsonObject() {{
-      put("identifier", "user1");
-      put("code", "249916");
-    }};
+    JsonObject credentials = new JsonObject()
+      .put("identifier", "user1")
+      .put("code", "249916");
+
     authProvider.authenticate(credentials);
   }
 }
--- a/vertx-auth-otp/src/main/java/examples/hotp/HotpAuthExamples.java
+++ b/vertx-auth-otp/src/main/java/examples/hotp/HotpAuthExamples.java
@@ -45,7 +45,7 @@ public class HotpAuthExamples {
   }
 
   public static void example2() {
-    final HotpAuthOptions hotpAuthOptions = new HotpAuthOptions(6, 0 ,0);
+    final HotpAuthOptions hotpAuthOptions = new HotpAuthOptions();
     final HotpAuth authProvider = HotpAuth.create(hotpAuthOptions);
 
     // key
--- a/vertx-auth-otp/src/main/java/examples/hotp/HotpAuthExamples.java
+++ b/vertx-auth-otp/src/main/java/examples/hotp/HotpAuthExamples.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2021 Dmitry Novikov
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package examples.hotp;
+
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.auth.impl.UserImpl;
+import io.vertx.ext.auth.otp.hotp.HotpAuth;
+import io.vertx.ext.auth.otp.hotp.HotpAuthOptions;
+import io.vertx.ext.auth.otp.utils.AuthenticatorUriGenerator;
+import io.vertx.ext.auth.otp.utils.OtpKey;
+import io.vertx.ext.auth.otp.utils.OtpKeyGenerator;
+import net.glxn.qrgen.core.image.ImageType;
+import net.glxn.qrgen.javase.QRCode;
+import org.apache.commons.codec.binary.Base32;
+
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.net.URI;
+
+public class HotpAuthExamples {
+
+  public static void example1() {
+    // generate key
+    OtpKeyGenerator keyGenerator = new OtpKeyGenerator();
+    OtpKey otpKey = keyGenerator.generate();
+
+    // create qr code
+    URI uri = AuthenticatorUriGenerator.generateHotpUri(otpKey, "vertx_hotp", "user_1", "service_1");
+    QRCode qrCode = QRCode.from(uri.toString()).to(ImageType.PNG).withSize(512, 512);
+
+    try(FileOutputStream fos = new FileOutputStream("otp_key.png")) {
+      qrCode.writeTo(fos);
+    } catch (IOException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+
+  public static void example2() {
+    final HotpAuthOptions hotpAuthOptions = new HotpAuthOptions(6, 0 ,0);
+    final HotpAuth authProvider = HotpAuth.create(hotpAuthOptions);
+
+    // key
+    final String userKey = "2P3V27SMNQHEE4CYK26ZHTN5HPIGWT4R";
+    final byte[] keyBytes = new Base32(false).decode(userKey);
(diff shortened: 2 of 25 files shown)


vertx-auth-otp/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+    <!-- https://mvnrepository.com/artifact/commons-codec/commons-codec -->
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+      <version>1.15</version>
+    </dependency>
+    <dependency>
+      <groupId>com.github.otricadziziusz</groupId>
+      <artifactId>java-otp</artifactId>
+      <version>generate_strings-SNAPSHOT</version>
```
pmlopes: This is a blocker for a final release, we can only release non snapshots

vertx-auth-otp/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+    </dependency>
+    <dependency>
```
pmlopes: Is this dependency always required? My assumption (which may be wrong) is that QR codes are not always required, in this case we could mark this dependency optional, specially when it is mostly used on a UI side, not strictly necessary this kind of module.

vertx-auth-otp/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+  <parent>
+    <artifactId>vertx-auth</artifactId>
+    <groupId>io.vertx</groupId>
+    <version>4.1.1-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>vertx-auth-otp</artifactId>
+
+  <repositories>
```
pmlopes: Current release rules with the eclipse foundation, prevent us from using other sources than maven central, can we workaround this?

vertx-auth-otp/src/main/java/examples/hotp/HotpAuthExamples.java
```diff
@@ -0,0 +1,71 @@
...
+  public static void example1() {
+    // generate key
+    OtpKeyGenerator keyGenerator = new OtpKeyGenerator();
+    OtpKey otpKey = keyGenerator.generate();
+
+    // create qr code
+    URI uri = AuthenticatorUriGenerator.generateHotpUri(otpKey, "vertx_hotp", "user_1", "service_1");
+    QRCode qrCode = QRCode.from(uri.toString()).to(ImageType.PNG).withSize(512, 512);
+
+    try(FileOutputStream fos = new FileOutputStream("otp_key.png")) {
```
pmlopes: This can be problematic for a vert.x application as `FileOutputStream` is a blocking operation, can we get the `byte[]` data directly? If we can we should wrap this functionality with the API and return a vert.x Buffer so it can be used with the asynchronous APIs.

vertx-auth-otp/src/main/java/examples/hotp/HotpAuthExamples.java
```diff
@@ -0,0 +1,71 @@
...
+    final HotpAuthOptions hotpAuthOptions = new HotpAuthOptions();
+    final HotpAuth authProvider = HotpAuth.create(hotpAuthOptions);
+
+    // key
+    final String userKey = "OK7JVNHJO5ZMC57QLYJ6QNTOZFKVN76Y";
+    final byte[] keyBytes = new Base32(false).decode(userKey);
+    final OtpKey otpKey = new OtpKey(keyBytes, "HmacSHA1");
+
+    // request hotp for user
+    JsonObject principal = new JsonObject() {{
```
pmlopes: `JsonObject` has a fluent API, so you don't need to do `{{ ... }}`. you can skip them.

vertx-auth-otp/src/main/java/examples/hotp/HotpAuthExamples.java
```diff
@@ -0,0 +1,71 @@
...
+      put("identifier", "user1");
+      put("key", otpKey.getBase32());
+      put("counter", 0);
+    }};
+    authProvider.requestHotp(new UserImpl(principal), userAsyncResult -> {});
```
pmlopes: The recommended way to use the API is to use the factory method in the public interfaces, so instead of `new UserImpl()` you should replace it with: `User.create()`

pmlopes, line 62: `requestHotp()` should be renamed with `authenticate()`, this way is adheres to the parent interface and we have a common API to perform authentication. Also if the user requires custom parameters, I think we should have a `Credentials` implementation, like for example in the `HTDigest` module.

vertx-auth-otp/src/main/java/examples/hotp/HotpAuthExamples.java
```diff
@@ -0,0 +1,71 @@
...
+
+    // auth user hotp
+    JsonObject credentials = new JsonObject() {{
```
pmlopes, line 65: Also if the user requires custom parameters, I think we should have a Credentials implementation, like for example in the HTDigest module.

vertx-auth-otp/src/main/java/io/vertx/ext/auth/otp/hotp/HotpAuth.java
```diff
@@ -0,0 +1,35 @@
...
+import io.vertx.core.Handler;
+import io.vertx.ext.auth.User;
+import io.vertx.ext.auth.authentication.AuthenticationProvider;
+import io.vertx.ext.auth.impl.UserImpl;
+import io.vertx.ext.auth.otp.hotp.impl.HotpAuthImpl;
+
+@VertxGen
+public interface HotpAuth extends AuthenticationProvider {
+
+  @GenIgnore
```
pmlopes: You should not `GenIgnore`, otherwise other languages (e.g.: Kotlin, Groovy, RX, JavaScript) will not see this method. You can solve it, by using `User` in the param instead of `UserImpl`.

vertx-auth-otp/src/main/java/io/vertx/ext/auth/otp/hotp/HotpAuth.java
```diff
@@ -0,0 +1,35 @@
...
+  void requestHotp(UserImpl user, Handler<AsyncResult<User>> resultHandler);
+
+  @GenIgnore
```
pmlopes: Same comment as in the previous method.

pmlopes:
@otricadziziusz great contribution! Your PR is in very good shape, there are a few comments added. My main concern atm is the use of non released dependencies to maven central which can be problematic to adhere to the eclipse foundation release process, plus the generation of QR codes, seems to be a blocking task which can be a problem.

Perhaps we should move the QR code to a vert.x web handler instead? Finally the use of BASE32 as byte[] isn't a big deal per se, but we probably should hide it behind a vert.x buffer so we can interop with all the code generators and languages.

otricadziziusz:
@pmlopes Thanks for review.

About a version of dependynce. The author of the library with OTP generator for now has not release the version with the desired feature, I have fork a repo and explicitly indicated the branch with the desired feature. Yes, now I understand what could be the problem. I could replace this dependency and just use reference implementation from RFC. I think this will be true solution.

I agree with the comment about the QR code, now I think about remove generation QR code. After all, there is only a URI inside it, which is generated through an io.vertx.ext.auth.otp.utils.AuthenticatorUriGenerator. URI will be enough to get the QR code with any convenient library.

About HotpAuth.requestHotp, it is now used as a supplier critical information 
for check otp challenge. I think about mechanism supply information for auth and revoke request otp authentication.

I am starting to work on your comments and perhaps in the process I will contact you here with questions or discussion?

otricadziziusz:
@pmlopes, I am finished working on comments, i replaced the dependency with the reference implementation from RFC4226. But, I need a hint about buffer. As I understand it, I need to change the encoding base32 for key on user. I am unable to place it in json due to io.vertx.core.json.DecodeException. I think that now, except use encoding, one cannot put key in json.

pmlopes:
@otricadziziusz thanks! I'll have a new look tomorrow in the morning... In the meantime it would be great to see how this module would work with vertx-web, perhaps a small helper handler could be created to perform 2FA tasks? Say return the QR code PNG, validate code (after user it authenticated) etc...

otricadziziusz:
Yes of course! I will do it shortly.

vertx-auth-otp/src/main/java/org/openauthentication/otp/OneTimePasswordAlgorithm.java
```diff
@@ -0,0 +1,187 @@
+package org.openauthentication.otp;
```
pmlopes: To keep the generated final jar compatible with JPMS we must hold all code under the module package, e.g.:

```
package io.vertx.ext.auth.otp.impl.org.openauthentication.otp;
```

vertx-auth-otp/src/main/java/io/vertx/ext/auth/otp/utils/AuthenticatorUriGenerator.java
```diff
@@ -0,0 +1,24 @@
...
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.ext.auth.otp.utils;
+
+import java.net.URI;
+
+public class AuthenticatorUriGenerator {
```
pmlopes: We should annotate this with `@VertxGen` this would allow it to be used from polyglot contexts. This also means that the code should be slightly different:

```
@VertxGen
public interface AuthenticatorUriGenerator {
```

vertx-auth-otp/src/main/java/io/vertx/ext/auth/otp/utils/AuthenticatorUriGenerator.java
```diff
@@ -0,0 +1,24 @@
...
+
+  private static final String HOTP_GOOGLE_AUTH_URI_TEMPLATE = "otpauth://hotp/%s:%s?secret=%s&issuer=%s";
+
+  public static URI generateHotpUri(OtpKey otpKey, String label, String user, String issuer) {
```
pmlopes: Making it polyglot will cause problems with `URI` so it would require a `@GenIgnore(GenIgnore.PERMITED_TYPE)`

vertx-auth-otp/src/main/java/io/vertx/ext/auth/otp/utils/OtpKey.java
```diff
@@ -0,0 +1,53 @@
...
+
+package io.vertx.ext.auth.otp.utils;
+
+import org.apache.commons.codec.binary.Base32;
+
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
+import java.security.Key;
+
+public class OtpKey implements SecretKey {
```
pmlopes: This one also should be transformed to:

```
@VertxGen
public interface OtpKey {
```

And the implementation then can implement both `OtpKey` and `SecretKey` as keeping `SecretKey` in the top level interface would fail the generation.

vertx-auth-otp/src/main/java/io/vertx/ext/auth/otp/utils/OtpKeyGenerator.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.ext.auth.otp.utils;
+
+import javax.crypto.KeyGenerator;
+import java.security.NoSuchAlgorithmException;
+
+public class OtpKeyGenerator {
```
pmlopes: Same `@VertxGen`.

Basically all public APIs should be annotated, anything that is implementation should go under a `impl` package and there there's no need to annotate.

vertx-auth-otp/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+  <properties>
+    <doc.skip>false</doc.skip>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-auth-common</artifactId>
+    </dependency>
+    <dependency>
```
pmlopes, line 38: Instead of the core test dependency, you should use:
```xml
<dependency>
      <groupId>io.vertx</groupId>
      <artifactId>vertx-unit</artifactId>
      <scope>test</scope>
    </dependency>
``` 

Which means that tests should be slightly changed later on

vertx-auth-otp/src/test/java/io/vertx/ext/auth/otp/HotpAuthOptionsTest.java
```diff
@@ -0,0 +1,127 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.ext.auth.otp;
+
+import io.vertx.ext.auth.otp.hotp.HotpAuthOptions;
+import io.vertx.test.core.VertxTestBase;
+import org.junit.Test;
+
+public class HotpAuthOptionsTest extends VertxTestBase {
```
pmlopes: Comming from the dependency change in the pom you can just do:

```java
@RunWith(VertxUnitRunner.class)
public class HotpAuthOptionsTest {

  @Rule
  public RunTestOnContext rule = new RunTestOnContext();

  @Test
  public void test1(TestContext should) {
...
```

If you test requires asynchronous calls your 1st line in the test is:

```java
 final Async test = should.async();
```

So you can now do all assertions against the `should` object, and signal that you're done with `test.complete()`.

pmlopes (changes requested):
I've a few comments. If you like and it would make it easier, I could merge the PR as is into a seperate branch, so I could also contribute with the small fixes, instead of asking you to do all the work.

I can imagine that you would want to do it all by yourself, but if you want some help, don't fell shy to let me know and we can work together that way!

otricadziziusz:
Thanks! I made changes that separate public api and implementation.

I agree, if it is more convenient for us then we could merge the PR into a separate branch. Yes, I would not mind your help.

otricadziziusz:
@pmlopes I am want starting to do examples work with module and would like to find out. In what form would it be more convenient to present an example of work module with vertx-web?

pmlopes:
@otricadziziusz I think the best way to do this is first we create a PoC of a webapp (nothing fancy), mostly get the basic concepts working, say for example:

1. create a properties auth with a test user
2. a form to register otp + the qr code
3. a confirm creation page where the 1st code is tested and the necessary data is stored (memory for now)
4. a login page that use the properties initially with a simple form
5. a second page where after login you enter the generated code and we can validate

Once we have all this working we can see how to isolate the OTP web handler code into a single class. WDYT?

pmlopes:
@otricadziziusz I've created the following test app:

```java
package org.acme;

import io.vertx.core.AbstractVerticle;
import io.vertx.core.Vertx;
import io.vertx.core.buffer.Buffer;
import io.vertx.core.json.JsonObject;
import io.vertx.ext.auth.User;
import io.vertx.ext.auth.otp.AuthenticatorUriGenerator;
import io.vertx.ext.auth.otp.OtpKey;
import io.vertx.ext.auth.otp.OtpKeyGenerator;
import io.vertx.ext.auth.otp.hotp.HotpAuth;
import io.vertx.ext.auth.otp.hotp.HotpAuthOptions;
import io.vertx.ext.auth.otp.hotp.HotpCredentials;
import io.vertx.ext.web.Router;
import io.vertx.ext.web.handler.BodyHandler;

import net.glxn.qrgen.core.image.ImageType;
import net.glxn.qrgen.javase.QRCode;

import java.net.URI;
import java.util.HashMap;
import java.util.Map;

public class Foo extends AbstractVerticle {

    static class DBEntry {
        final String id;
        final OtpKey key;
        int counter;

        DBEntry(String id, OtpKey key) {
            this.id = id;
            this.key = key;
        }
    }

    public static void main(String[] args) {
        Vertx.vertx().deployVerticle(new Foo());
    }

    private final Map<String, DBEntry> DB = new HashMap<>();

    private static final String ID = "Paulo";

    @Override
    public void start() {

        OtpKeyGenerator keyGen = OtpKeyGenerator.create();
        HotpAuth hotp = HotpAuth.create(new HotpAuthOptions());

        Router app = Router.router(vertx);
        // for form data parsing
        app.route().handler(BodyHandler.create());

        app.get("/register")
                .handler(ctx -> {
                    OtpKey otpKey = keyGen.generate();

                    // Create user in the database
                    DB.put(ID, new DBEntry(ID, otpKey));

                    URI uri = AuthenticatorUriGenerator.create().generateHotpUri(otpKey, "vertx_hotp", "user_1", "service_1");

                    ctx.response()
                            .putHeader("Content-Type", "image/png")
                            .end(
                                    Buffer.buffer(
                                            QRCode
                                                    .from(uri.toString())
                                                    .to(ImageType.PNG)
                                                    .withSize(512, 512)
                                                    .stream()
                                                    .toByteArray()));

                });

        app.post("/verify")
                .handler(ctx -> {
                    String token = ctx.getBodyAsJson().getString("token");

                    DBEntry entry = DB.get(ID);

                    JsonObject principal = new JsonObject()
                            .put("identifier", entry.id)
                            .put("key", entry.key.getBase32())
                            .put("counter", entry.counter);

                    hotp.requestHotp(User.create(principal), requestHotp -> {
                        if (requestHotp.failed()) {
                            ctx.fail(requestHotp.cause());
                        } else {
                            hotp.authenticate(new HotpCredentials(ID, token))
                                    .onFailure(ctx::fail)
                                    .onSuccess(user -> {
                                        // Update user data
                                        ctx.json(new JsonObject().put("verified", true));
                                    });
                        }
                    });
                });

        vertx.createHttpServer()
                .requestHandler(app)
                .listen(8080)
                .onFailure(Throwable::printStackTrace)
                .onSuccess(server -> {
                    System.out.println("Server listening: http://localhost:" + server.actualPort());
                });
    }

}
```

Can you check if this is correct? I can generate a QR code but the codes my app generates are always "invalid".

otricadziziusz:
@pmlopes Thanks for creating the example! I checked this example, the only first code works for me. Since the current example does not save information about the counter to the database and it is always 0. All following generated codes will be invalid.

pmlopes:
@otricadziziusz which app do you use ? I was trying to use freeotp but even the first code fails. Must I use Google authenticator instead?

otricadziziusz:
```
package org.acme;

import io.vertx.core.AbstractVerticle;
import io.vertx.core.Vertx;
import io.vertx.core.buffer.Buffer;
import io.vertx.core.json.JsonObject;
import io.vertx.ext.auth.User;
import io.vertx.ext.auth.otp.AuthenticatorUriGenerator;
import io.vertx.ext.auth.otp.OtpKey;
import io.vertx.ext.auth.otp.OtpKeyGenerator;
import io.vertx.ext.auth.otp.hotp.HotpAuth;
import io.vertx.ext.auth.otp.hotp.HotpAuthOptions;
import io.vertx.ext.auth.otp.hotp.HotpCredentials;
import io.vertx.ext.web.Router;
import io.vertx.ext.web.handler.BodyHandler;

import net.glxn.qrgen.core.image.ImageType;
import net.glxn.qrgen.javase.QRCode;

import java.net.URI;
import java.util.HashMap;
import java.util.Map;

public class Foo extends AbstractVerticle {

  static class DBEntry {
    final String id;
    final OtpKey key;
    int counter;

    DBEntry(String id, OtpKey key) {
      this.id = id;
      this.key = key;
    }
  }

  public static void main(String[] args) {
    Vertx.vertx().deployVerticle(new Foo());
  }

  private final Map<String, DBEntry> DB = new HashMap<>();

  private static final String ID = "Paulo";

  @Override
  public void start() {

    OtpKeyGenerator keyGen = OtpKeyGenerator.create();
    HotpAuth hotp = HotpAuth.create(new HotpAuthOptions());

    Router app = Router.router(vertx);
    // for form data parsing
    app.route().handler(BodyHandler.create());

    app.get("/register")
      .handler(ctx -> {
        OtpKey otpKey = keyGen.generate();

        System.out.println(otpKey.getBase32());
        // Create user in the database
        DB.put(ID, new DBEntry(ID, otpKey));

        URI uri = AuthenticatorUriGenerator.create().generateHotpUri(otpKey, "vertx_hotp", "user_1", "service_1");

        ctx.response()
          .putHeader("Content-Type", "image/png")
          .end(
            Buffer.buffer(
              QRCode
                .from(uri.toString())
                .to(ImageType.PNG)
                .withSize(512, 512)
                .stream()
                .toByteArray()));

      });

    app.post("/verify")
      .handler(ctx -> {
        String token = ctx.getBodyAsJson().getString("token");

        DBEntry entry = DB.get(ID);

        JsonObject principal = new JsonObject()
          .put("identifier", entry.id)
          .put("key", entry.key.getBase32())
          .put("counter", entry.counter);

        hotp.requestHotp(User.create(principal), requestHotp -> {
          if (requestHotp.failed()) {
            ctx.fail(requestHotp.cause());
          } else {
            hotp.authenticate(new HotpCredentials(ID, token))
              .onFailure(ctx::fail)
              .onSuccess(user -> {
                // Update user data
                entry.counter = user.attributes().getInteger("counter");
                DB.put(ID, entry);
                ctx.json(new JsonObject().put("verified", true));
              });
          }
        });
      });

    vertx.createHttpServer()
      .requestHandler(app)
      .listen(8080)
      .onFailure(Throwable::printStackTrace)
      .onSuccess(server -> {
        System.out.println("Server listening: http://localhost:" + server.actualPort());
      });
  }
}
```

@pmlopes Made changes to store the counter upon successful authentication. Now all my generated codes are working correctly. Could you try some more with current changes.

otricadziziusz:
@pmlopes Yes, to test the implementation, I used Google Authenticator.

pmlopes:
Closing as the work continued on #490 and #491 and it has been merged.

PR #491, opened 2021-06-29
Feature/otp/totp

otricadziziusz:
Motivation:

I continue to work on the module otp, added totp implementation


otricadziziusz:
```
package org.acme;

import io.vertx.core.AbstractVerticle;
import io.vertx.core.Future;
import io.vertx.core.Vertx;
import io.vertx.core.buffer.Buffer;
import io.vertx.core.json.JsonObject;
import io.vertx.ext.auth.otp.Authenticator;
import io.vertx.ext.auth.otp.OtpKey;
import io.vertx.ext.auth.otp.OtpKeyGenerator;
import io.vertx.ext.auth.otp.totp.TotpAuth;
import io.vertx.ext.auth.otp.totp.TotpCredentials;
import io.vertx.ext.web.Router;
import io.vertx.ext.web.handler.BodyHandler;

import net.glxn.qrgen.core.image.ImageType;
import net.glxn.qrgen.javase.QRCode;

import java.util.Map;
import java.util.concurrent.ConcurrentHashMap;

public class Foo extends AbstractVerticle {

  private final AuthenticatorDataBase DB = new AuthenticatorDataBase();

  public static void main(String[] args) {
    Vertx.vertx().deployVerticle(new Foo());
  }

  private static final String ID = "Paulo";

  @Override
  public void start() {

    OtpKeyGenerator keyGen = OtpKeyGenerator.create();
    TotpAuth totp = TotpAuth.create()
      .authenticatorFetcher(DB::fetch)
      .authenticatorUpdater(DB::upsert);

    Router app = Router.router(vertx);
    // for form data parsing
    app.route().handler(BodyHandler.create());

    app.get("/register")
      .handler(ctx -> {
        OtpKey otpKey = keyGen.generate();

        System.out.println(otpKey.getKey());
        // Create user in the database
        final Authenticator authenticator = new Authenticator(true)
          .setIdentifier(ID)
          .setKey(otpKey.getKey())
          .setAlgorithm(otpKey.getAlgorithm())
          .setCounter(0);

        DB.upsert(authenticator)
          .onFailure(ctx::fail)
          .onSuccess(v -> ctx.response()
            .putHeader("Content-Type", "image/png")
            .end(
              Buffer.buffer(
                QRCode
                  .from(totp.generateUri(otpKey, authenticator.getCounter(), "service_1", ID))
                  .to(ImageType.PNG)
                  .withSize(512, 512)
                  .stream()
                  .toByteArray())));
      });

    app.post("/verify")
      .handler(ctx -> {
        String code = ctx.getBodyAsJson().getString("code");

        totp.authenticate(new TotpCredentials(ID, code))
          .onFailure(err -> ctx.fail(401, err))
          .onSuccess(user -> ctx.json(new JsonObject().put("verified", true)));
      });

    vertx.createHttpServer()
      .requestHandler(app)
      .listen(8080)
      .onFailure(Throwable::printStackTrace)
      .onSuccess(server -> {
        System.out.println("Server listening: http://localhost:" + server.actualPort());
      });
  }

  public static class AuthenticatorDataBase {

    private final Map<String, Authenticator> DB = new ConcurrentHashMap<>();

    public Future<Authenticator> fetch(String id) {
      if (DB.containsKey(id)) {
        return Future.succeededFuture(DB.get(id));
      } else {
        return Future.succeededFuture();
      }
    }

    public Future<Void> upsert(Authenticator authenticator) {
      DB.put(authenticator.getIdentifier(), authenticator);
      return Future.succeededFuture();
    }
  }
}

```

@pmlopes I modified your example to check totp auth.
Implementation of authenticators is very similar and differs only in working with a counter. I made them as isolated as possible and did not generalize.

pmlopes:
@otricadziziusz this looks great. All we need now is a basic manual + fill the javadocs.

One note I see from the examples we should improve:

In the example we have:

```java
OtpKey otpKey = keyGen.generate();
// Create user in the database
final Authenticator authenticator = new Authenticator(true)
  .setIdentifier(ID)
  .setKey(otpKey.getKey())
  .setAlgorithm(otpKey.getAlgorithm())
  .setCounter(0);

DB.upsert(authenticator)
  .onFailure(ctx::fail)
  .onSuccess(v -> ...
```

In a way, this isn't correct IMHO as it forces users to know the internal details of the implementation regarding the authenticator load and store. I think that we should add one extra method to `Hotp`/`Totp` public interfaces:

```java
  Future<Authenticator> createAuthenticator(String id, OtpKey otpKey) {
    // Create user in the database
    final Authenticator authenticator = new Authenticator(true)
      .setIdentifier(id)
      .setKey(otpKey.getKey())
      .setAlgorithm(otpKey.getAlgorithm())
      .setCounter(0);

    return updater
      .apply(authenticator)
      .map(authenticator);
  }
```

This way, users will not have to deal with the internals of how authenticators are fetched or updated.

For the manual, you can use this as inspiration: https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-htpasswd/src/main/asciidoc/index.adoc

Just a simple asciidoc explaining the core functionality and an example or two. Note that the examples should refer to a Java class so if the code changes, the docs will change automatically too.

Also, the example should be low level in this module, by this I mean no vertx-web involved, so basically all it needs to cover is how to register authenticators, how to validate codes.

From here we can now start on vertx-web, where we can add a handler that can reduce the boilerplate code.

pmlopes:
@otricadziziusz great, I'll try to review it tomorrow and get it ready for merge so we will make the 4.2.0 release as tech preview

pmlopes:
Thank you @otricadziziusz, really a great work! 👍

otricadziziusz:
Thank you @pmlopes for all code reviews! I will definitely still participate in vertx!

pmlopes:
I've created an new issue in vertx web so we can track a helper handler to create qr codes and manage the registration / validation

PR #496, opened 2021-07-07
Fix algorithm in example 18

re-thc:
Should be ES instead of RS as the rest of the example is for ES.

Motivation:

Fix docs


pmlopes:
Thanks!

Issue #506, opened 2021-09-23
MongoAuth 4.1.x does not use custom property names correctly


sophokles73:
### Version

MongoAuth 4.1.4

### Context

When setting the *usernameCredentialsField* and *passwordCredentialsField* configuration properties on class MongoAuthenticationOptions to custom values, then the authentication of a credentials JsonObject using the configured property names fails.

### Do you have a reproducer?

No, but I have a PR fixing the issue.

pmlopes:
Thanks!

PR #508, opened 2021-09-24
Use configured custom property names

sophokles73:
This is for #506
 
The MongoAuthenticationImpl class did not properly consider the (custom)
field names set on MongoAuthenticationOptions.


pmlopes:
Thanks! :+1:

Issue #464, opened 2021-03-11
User object should have a subject() read-only getter


pmlopes:
This method is useful to ensure that users can be identified uniquely.

A use case for this feature would be during logging we could log the username just like Apache does.

pmlopes:
Update, instead of "name" we probably should call it `subject()`

There are a couple of locations where this information can be found:

* `principal.username`: usually for username/password authentication or webauthn
* `attributes.idToken.sub`: for OIDC
* `attributes.accessToken.sub`: for Oauth2
* `attributes.accessToken.sub` + `attributes.accessToken[.act]+.sub`: for TokenExchange with Oauth2 (can be recursive)
* `attributes.accessToken.sub` + `attributes.accessToken.may_act.sub`: for TokenExchange with Oauth2
* webauthn could also pass the `userHandle` in the initial credentials request we should keep track of it too

When, dealing with multiple subjects (impersonation/delegation) the method should return a concatenation of the subjects using a delimiter, for example `+` or `/`.

PR #518, opened 2021-11-29
Initial definition of the Subject interface

pmlopes:

Motivation:

Fixes issue #464


pmlopes:
This idea has been dropped for now as it was complex. Instead we allow impersonation at vertx-web using a small swapping mechanism.

Issue #501, opened 2021-08-16
Improve OIDC/OAuth2 Flow discovery


pendula95:
#### Describe the feature

As explained here we should improve the discovery mechanism to support all possible flows enabled by the provider.
We should not force user to choose the flow when creating a provider but check if the request is possible when request is made.
Store possible flows in in a list.

https://github.com/vert-x3/vertx-web/pull/2016#issuecomment-895141723

pendula95:
is this still up for taking?

pmlopes:
@pendula95 yes, I did try to see if quickly remove the single entry would be trivial, but the assumption that there is a single flow is used internally in several switch statements which will require some real refactor.

If you're interested in working on this, lets start a discussion here on the required changes and you can start a pull request we can work together

pendula95:
If I am interpreting this right the first step would be to remove `OAuth2Options setFlow(OAuth2FlowType flow)` from `OAuth2Options` and flow should be passed in `authenticate(JsonObject credentials)` from `AuthenticationProvider` or even to be automatically concluded by the content of credentials?

pmlopes:
Yes.

* https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/Oauth2Credentials.java should have a new field to specify the flow
* https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2Options.java should deprecate setting the flow

With those in place, start replacing all usages of the deprecated getter with the getter from the credentials object.

A second action is to start working on the discovery feature updates. During the discovery, we should collect the allowed flows. Then during the calls https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2API.java ensure that the flow in the credentials is permitted.

pmlopes:
@pendula95 I believe this is wrong:

https://github.com/vert-x3/vertx-auth/blob/04ac36df802b8c09d4a15400fb5885bc092ad7ab/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java#L304-L316

[...]

PR #529, opened 2021-12-17
Updating authenticate to expect OAuth2Credentials for OBO calls

pmlopes:
Initial branch to implement #501

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2FlowType.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2FlowType.java
@@ -69,7 +69,7 @@ public enum OAuth2FlowType {
    * the security implications of using implicit grants, especially when the
    * authorization code grant type is available.
    */
-  IMPLICIT(null),
+  IMPLICIT("implicit"),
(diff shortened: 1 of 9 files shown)


pmlopes:
Don't forget to add a note to:

OBO calls, require an `OAuth2Credentials` shape object to authenticate, not a `TokenCredentials`. Using `OAuth2Credentials` signals the start of the oauth2 flow, while using `TokenCredentials` signals validation/introspection.

The original behavior was incorrect, and can be seen as an implementation bug.

pmlopes:
We probably should update the enum to include more values, and we need to discuss how to handle this ackwardness:

https://github.com/vert-x3/vertx-auth/blob/a158def51e7b14504089a9d547e944a760814cd0/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2FlowType.java#L105-L115

pendula95:
I will try and obtain all possible values and fill the enum. On the other hand I don't see the need to add more values to it as we don't support flows that are currently not in enum. Adding more values only makes sense if we are going to add implementation for them. Regarding AAD_OBO and AUTH_JWT I had not exposure to it so I have little knowledge about it.Should read some RFCs.

PR #543, opened 2022-03-12
add gitee oauth provider

oschina:
Motivation:

Gitee is the largest open source code hosting platform in China, with more than 8 million registered developers, and many developers in mainland China use Gitee as a third-party login platform.

I very much hope that Vert.x can officially add OAuth authentication support for Gitee.

Conformance:

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/providers/GiteeAuth.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/providers/GiteeAuth.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.ext.auth.oauth2.providers;
+
+import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.core.Vertx;
+import io.vertx.core.http.HttpClientOptions;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.auth.oauth2.OAuth2Auth;
+import io.vertx.ext.auth.oauth2.OAuth2FlowType;
+import io.vertx.ext.auth.oauth2.OAuth2Options;
+
+/**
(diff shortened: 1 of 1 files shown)


gaol:
Cool !

Could you please[ sign ECA](https://www.eclipse.org/legal/ECA.php) and squash your commits into one for the review ? thanks :)

oschina:
> Cool !
> 
> Could you please[ sign ECA](https://www.eclipse.org/legal/ECA.php) and squash your commits into one for the review ? thanks :)

ECA signed and commit squashed :)

pmlopes:
@oschina, just a quick question. Does `gitee` support openid auto discovery? If it does it would be nice to also support it as it makes configuration even easier for end users.

For example here's what, Azure does:

https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration

If you have such URL, vert.x already knows how to parse this and you can do:

```java
/**
   * Create a OAuth2Auth provider for OpenID Connect Discovery. The discovery will use the default site in the
   * configuration options and attempt to load the well known descriptor. If a site is provided (for example when
   * running on a custom instance) that site will be used to do the lookup.
   * <p>
   * If the discovered config includes a json web key url, it will be also fetched and the JWKs will be loaded
   * into the OAuth provider so tokens can be decoded.
   *
   * @param vertx   the vertx instance
   * @param config  the initial config
   * @param handler the instantiated Oauth2 provider instance handler
   */
  static void discover(final Vertx vertx, final OAuth2Options config, final Handler<AsyncResult<OAuth2Auth>> handler) {
    // don't override if already set
    final String site = config.getSite() == null ? "https://gitee.com/<base to .well-known dir>" : config.getSite(); // <--- HERE

    OpenIDConnectAuth.discover(
      vertx,
      new OAuth2Options(config)
        .setSite(site)
      handler);
  }

  /**
   * Create a OAuth2Auth provider for OpenID Connect Discovery. The discovery will use the default site in the
   * configuration options and attempt to load the well known descriptor. If a site is provided (for example when
   * running on a custom instance) that site will be used to do the lookup.
   * <p>
   * If the discovered config includes a json web key url, it will be also fetched and the JWKs will be loaded
   * into the OAuth provider so tokens can be decoded.
   *
   * @see GoogleAuth#discover(Vertx, OAuth2Options, Handler)
   * @param vertx   the vertx instance
   * @param config  the initial config
   * @return future with the instantiated Oauth2 provider instance handler
   */
  static Future<OAuth2Auth> discover(final Vertx vertx, final OAuth2Options config) {
    Promise<OAuth2Auth> promise = Promise.promise();
    discover(vertx, config, promise);
    return promise.future();
  }
```

oschina:
> @oschina, just a quick question. Does `gitee` support openid auto discovery? If it does it would be nice to also support it as it makes configuration even easier for end users.

@pmlopes  Unfortunately, Gitee.com does not currently support OpenID, we will evaluate this and support this feature in subsequent iterations.

PR #545, opened 2022-03-13
Use passwordProtection to read MAC keys from keystore

runtarinn:
Use the passwordProtection object if not null to supply keys for MAC keys instead of using the keystore password all the time. This is consistent with how asymmetric keys password is supplied.


pmlopes:
Thanks @runtarinn !

PR #546, opened 2022-03-16
Ensure that the PRNG works with and without context

pmlopes:
Motivation:

Downstream projects (Quarkus) block access to the vert.x context. Even though PRNG was built to avoid duplication, we cannot throw exceptions as it would block applications from starting.

The behavior is changed to allow running without a context and fall back to having multiple instances.


cescoffier (approved):

FroMage:
Can we backport this?

pmlopes:
@FroMage cherry picked to 4.2.6

FroMage:
Great!

PR #549, opened 2022-04-05
Replace JDBC with SQL Client

pmlopes:
Motivation:

Currently we have 2 ways of achieving the same result. This PR shows that we can achive the same result just with one module.

If we all agree we, can deprecate the `vertx-auth-jdbc` like shiro.

### DIFF
--- a/vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/JDBCAuthentication.java
+++ b/vertx-auth-jdbc/src/main/java/io/vertx/ext/auth/jdbc/JDBCAuthentication.java
@@ -24,6 +24,8 @@ import io.vertx.ext.jdbc.JDBCClient;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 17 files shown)


pmlopes:
The current `vertx-jdbc-client` is capable of working with the same API as the remaining asynchronous clients. We now have 2 modules in auth that provide the same functionality:

* vertx-auth-jdbc
* vertx-auth-sql-client

I'd like to get the feeling if anyone opposes the idea of **deprecating** `vertx-auth-jdbc`, like we did in the past for `shiro`.

**NOTE**: This is only on the auth project, `jdbc-client` isn't going away!

Why would I like to deprecate `auth-jdbc`?

Because `auth-jdbc` is based on the previous API that isn't type safe. This API causes troubles with procedures and callable statements (on some drivers). Also, keeping it keeps the APIs inconsistent as we now have 2 different ways of achieving the same result, which is always a pattern of bad design I'd like to correct.

PR #563, opened 2022-05-31
Avoid explicitly depend on databind

pmlopes:

Motivation:

Internally we depend on jackson-cbor which requires databind. The CBOR decoding is not public and only used internally, so we could skip pulling the databind dependency as it has been a source of emergency releases due to CVEs.


pmlopes:
FYI: @paulgallagher75

Issue #567, opened 2022-06-24
4.2.1: NoSuchAlgorithmException: RSA-OAEP no longer a warning


rzorzorzo:
### Questions

Do not use this issue tracker to ask questions, instead use one of these [channels](https://vertx.io/community/). Questions will likely be closed without notice.

### Version

4.2.1

### Context

I encountered an exception which looks suspicious while ...

### Do you have a reproducer?

JWTUtil.setupJwtAuth();

throws an exception and does not iterate through all algorithms.

### Extra
i am migrating from 3.2.1 to 4.2.1
with 3.2.1 JWTUtil.setupJwtAuth(); did not throw an exception, so setup completed correctly.

to resolve change:

in io.vertx.ext.auth.jwt.impl.JWTAuthProviderImpl

        for (JsonObject jwk : jwks) {

            //this.jwt.addJWK(new JWK(jwk));

             // +++ should be warning not throw an exception

             try

             {

          this.jwt.addJWK(new JWK(jwk));

             }

             catch (Exception ex)

             {

                   System.out.println("warning: "+ex.getMessage());

             }

             // +++

        }

pmlopes:
Closed by the PR listed above

PR #569, opened 2022-06-27
Issues/567

pmlopes:
Motivation:

Keep the behavior consistent with previous releases and across modules. Unsupported JWKs are ignored.


pmlopes:
Fixes #567

PR #572, opened 2022-07-11
JWK max age property added

Vorimo:
Motivation: https://github.com/vert-x3/vertx-auth/issues/571

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2Options.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2Options.java
@@ -51,7 +51,8 @@ public class OAuth2Options {
   private static final String SCOPE_SEPARATOR = " ";
   private static final boolean VALIDATE_ISSUER = true;
   private static final boolean ROTATE_JWKS = true;
-  private static final long JWK_DEFAULT_AGE = 1L;
+  //seconds of JWK's default age (5 mins)
+  private static final long JWK_DEFAULT_AGE = 300L;
 
   private OAuth2FlowType flow;
   private List<String> supportedGrantTypes;
@@ -72,7 +73,7 @@ public class OAuth2Options {
   @Deprecated
   private boolean rotateJWKs;
   //seconds of JWKs lifetime
-  private long jwkMaxAge;
+  private long jwkMaxAgeInSeconds;
   // OpenID non standard
   private String tenant;
 
@@ -149,7 +150,7 @@ public class OAuth2Options {
     }
     jwkPath = other.getJwkPath();
     rotateJWKs = other.isRotateJWKs();
-    jwkMaxAge = other.getJwkMaxAge();
+    jwkMaxAgeInSeconds = other.getJwkMaxAgeInSeconds();
     httpClientOptions = other.getHttpClientOptions();
     userAgent = other.getUserAgent();
     supportedGrantTypes = other.getSupportedGrantTypes();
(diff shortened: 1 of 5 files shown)


vietj:
@pmlopes can you revi'ew this PR ?

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2Options.java
```diff
@@ -51,6 +51,7 @@ public class OAuth2Options {
   private static final String SCOPE_SEPARATOR = " ";
   private static final boolean VALIDATE_ISSUER = true;
   private static final boolean ROTATE_JWKS = true;
+  private static final long JWK_DEFAULT_AGE = 1L;
```
pmlopes: Shouldn't this be `-1`? Disabled by default

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2Options.java
```diff
@@ -68,7 +69,10 @@ public class OAuth2Options {
   private String introspectionPath;
   // JWK path RFC7517
   private String jwkPath;
+  @Deprecated
   private boolean rotateJWKs;
+  //seconds of JWKs lifetime
```
pmlopes, line 75: I believe all other time related config values are millisecond based, we should keep it consistent, but I'm not 100% sure about this

pmlopes (changes requested):
Can we confirm that we always use millis in config and change the docs and logic to expect always millis?

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2Options.java
Vorimo: Initially we discussed it to be 1, but if you think so, let's do -1

Vorimo:
@pmlopes it could be done if we are gonna expect `maxAge` header as millis, since we have `leeway` in millis. Or I can convert `maxAge` to millis and expect a new property in millis already

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2Options.java
pmlopes: `1` is too short, depending on the unit, it would mean that we would have a repeating timer every `sec/millisecond` refreshing keys. So usually we use `-1` for disabled, and any positive value for enable, although 0 and 1 are very small values for any one...

pmlopes:
@Vorimo yes, you're right! Let's use `seconds` as that's the unit used byt JWTOptions and user leeway. I think we should have an setter like:

```java
public Oauth2Options setJwkMaxAgeInSeconds(int seconds) { ... }
```

Like we do in `JWTOptions`: 

https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-common/src/main/java/io/vertx/ext/auth/JWTOptions.java#L103-L111

As a second PR we could align the remaining timing references to seconds, or if explicitly needed to be millis make it part of the name of the methods.

https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-webauthn/src/main/java/io/vertx/ext/auth/webauthn/WebAuthnOptions.java#L364-L376

https://github.com/vert-x3/vertx-auth/blob/master/vertx-auth-otp/src/main/java/io/vertx/ext/auth/otp/totp/TotpAuthOptions.java#L30

to mention a few...

Vorimo:
@pmlopes done

pmlopes (approved):

PR #575, opened 2022-07-15
Time properties renamed

Vorimo:
Alignment the timing references to seconds, or if explicitly needed to be millis/minutes/etc. make it part of the name of the methods.


pmlopes:
@Vorimo I think you got the idea, but it's not what I was thinking. What I think would make more sense would be to explicit mention the unit on the getter/setter.

For example:

```java
// Before the PR
setExpires()

// After the PR
setExpiresInSeconds()
// One of the two depending on the case
setExpiresInMilliseconds()
```

Vorimo:
@pmlopes I think now I done it correctly

Issue #581, opened 2022-07-28
WebAuthn: Add userId property to Authenticator and WebAuthnCredentials


mnylen:
#### Describe the feature

Depends on fixing #580 first.

#### 1. Add userId property to relevant classes

* Add `userId` property to both `Authenticator` and `WebAuthnCredentials`.
* After successful WebAuthn authentication or credential registration, the  `userId` property from `Authenticator` should be included in the principal of returned `User`.

#### 2. Populating the user id

After the property is added to relevant classes, the `Authenticator` instances passed to `authenticatorFetcher` and `authenticatorUpdater` should include the user ID (once #580 is fixed) property:

* When fetching authenticators with a query:
   * In `WebAuthn#createCredentialOptions()`: copy the value from the `user` object passed in as parameter.
   * In `WebAuthn#authenticate()`: use the user id defined in `WebAuthnCredentials` passed in as parameter
* When inserting or updating authenticators:
    * In `WebAuthn#authenticate()` for webauthn.get: use the user id defined in `WebAuthnCredentials`
    * In `WebAuthn#authenticate()` for webauthn.create: use the user id from the authenticator found by credential id lookup

#### 3. Extra check in authentication

Additionally, the verification process should ensure that the `userHandle` returned in the `AuthenticationAssertionResponse` matches the `userId` property on the `Authenticator` found with credential id lookup.

The spec:

> If the user was not identified before the [authentication ceremony](https://www.w3.org/TR/webauthn/#authentication-ceremony) was initiated,
>    verify that response.[userHandle](https://www.w3.org/TR/webauthn/#dom-authenticatorassertionresponse-userhandle) is present, and that the user identified by this value is the owner of credentialSource.

#### Use cases

This would allow Relying Party to index authenticators in persistent storage based on a stable user ID (read: a primary key for user table). User names (emails, nicknames, etc.) often change during user's lifecycle, but IDs don't.

Currently it's not possible to do this without doing an extra lookup to find user ID based on username (as authenticatorUpdater). 

#### Contribution

Already have a PR in the works at https://github.com/mnylen/vertx-auth/pull/1/files

PR #582, opened 2022-08-06
WebAuthn: Add user id to PublicKeyCredentialsCreateOptions, Authenticator and WebAuthnCredentials (#580, #581)

### LINKED ISSUE #580: WebAuthn: createCredentialOptions() uses a random UUID as user handle always
### Version

4.3.2

### Context

The documentation at https://vertx.io/docs/vertx-auth-webauthn/java/#_registration suggests adding `id` in the `user` object. This is, to my best understanding, supposed to be the Relying Party's unique identifier for the user, or the `user handle`.

In the implementation, however, the `user.id` in the `PublicKeyCredentialCreationOptions` generated does not use the given user id information. Instead, it always generates a random UUID there.

The implication of this is that when browsers or applications use the `PublicKeyCredentialCreationOptions` in `navigator.credentials.create` call, it is possible to create and store multiple credentials for the same user (as the user id is now completely random) on the authenticator. This is confusing for the users.

Only work around this issue is to override the `user.id` path from the `JsonObject` returned, which feels somewhat dirty.

How to fix:

* If `id` is present in the `user` object, base64 url encode that as `user.id` (makes it easier for most use cases, because in most cases the user ID will have a simple String representation (such as a number or String)).

* Or if `rawId` is present in the `user` object, use that directly as `user.id`

### Extra

According to WebAuthn spec level 2, the `user.id` (or user handle) is (https://www.w3.org/TR/webauthn/#dom-publickeycredentialuserentity-id):

> id, of type [BufferSource](https://heycam.github.io/webidl/#BufferSource)
>    The [user handle](https://www.w3.org/TR/webauthn/#user-handle) of the user account entity. A [user handle](https://www.w3.org/TR/webauthn/#user-handle) is an opaque [byte sequence](https://infra.spec.whatwg.org/#byte-sequence) with a maximum size of 64 bytes, and is not meant to be displayed to the user.
>
>    To ensure secure operation, authentication and authorization decisions MUST be made on the basis of this [id](https://www.w3.org/TR/webauthn/#dom-publickeycredentialuserentity-id) member, not the [displayName](https://www.w3.org/TR/webauthn/#dom-publickeycredentialuserentity-displayname) nor [name](https://www.w3.org/TR/webauthn/#dom-publickeycredentialentity-name) members. See Section 6.1 of [[RFC8266]](https://www.w3.org/TR/webauthn/#biblio-rfc8266).
>
>    The [user handle](https://www.w3.org/TR/webauthn/#user-handle) MUST NOT contain personally identifying information about the user, such as a username or e-mail address; see [§ 14.6.1 User Handle Contents](https://www.w3.org/TR/webauthn/#sctn-user-handle-privacy) for details. The [user handle](https://www.w3.org/TR/webauthn/#user-handle) MUST NOT be empty, though it MAY be null.
>
>    Note: the [user handle](https://www.w3.org/TR/webauthn/#user-handle) ought not be a constant value across different accounts, even for [non-discoverable credentials](https://www.w3.org/TR/webauthn/#non-discoverable-credential), because some authenticators always create [discoverable credentials](https://www.w3.org/TR/webauthn/#discoverable-credential). Thus a constant [user handle](https://www.w3.org/TR/webauthn/#user-handle) would prevent a user from using such an authenticator with more than one account at the [Relying Party](https://www.w3.org/TR/webauthn/#relying-party).

mnylen:
Motivation:

Fixes #580 and #581

### DIFF
--- a/vertx-auth-webauthn/src/main/generated/io/vertx/ext/auth/webauthn/AuthenticatorConverter.java
+++ b/vertx-auth-webauthn/src/main/generated/io/vertx/ext/auth/webauthn/AuthenticatorConverter.java
@@ -60,6 +60,10 @@ public class AuthenticatorConverter {
             obj.setUserName((String)member.getValue());
           }
           break;
+        case "userId":
+          if (member.getValue() instanceof String) {
+            obj.setUserId((String)member.getValue());
+          }
       }
     }
   }
@@ -90,6 +94,9 @@ public class AuthenticatorConverter {
     }
     if (obj.getUserName() != null) {
       json.put("userName", obj.getUserName());
+    }
+    if (obj.getUserId() != null) {
+      json.put("userId", obj.getUserId());
     }
   }
 }
--- a/vertx-auth-webauthn/src/main/java/io/vertx/ext/auth/webauthn/impl/WebAuthnImpl.java
+++ b/vertx-auth-webauthn/src/main/java/io/vertx/ext/auth/webauthn/impl/WebAuthnImpl.java
@@ -159,9 +159,20 @@ public class WebAuthnImpl implements WebAuthn {
 
   @Override
   public Future<JsonObject> createCredentialsOptions(JsonObject user) {
+    String userId;
+
+    if (user.getString("id") != null) {
+      userId = user.getString("id");
+    } else if (user.getString("rawId") != null) {
+      // For backwards compatibility, allow using rawId in place of id. Should be removed in future.
+      userId = user.getString("rawId");
+    } else {
+      // For backwards compatibility, if both id and rawId is missing, use a random base64url encoded UUID
+      userId = uUIDtoBase64Url(UUID.randomUUID());
+    }
 
     return fetcher
-      .apply(new Authenticator().setUserName(user.getString("name")))
+      .apply(new Authenticator().setUserId(userId).setUserName(user.getString("name")))
       .map(authenticators -> {
         // empty structure with all required fields
         JsonObject json = new JsonObject()
@@ -177,10 +188,11 @@ public class WebAuthnImpl implements WebAuthn {
         putOpt(json.getJsonObject("rp"), "icon", options.getRelyingParty().getIcon());
 
         // put non null values for User
-        putOpt(json.getJsonObject("user"), "id", uUIDtoBase64Url(UUID.randomUUID()));
+        putOpt(json.getJsonObject("user"), "id", userId);
         putOpt(json.getJsonObject("user"), "name", user.getString("name"));
         putOpt(json.getJsonObject("user"), "displayName", user.getString("displayName"));
         putOpt(json.getJsonObject("user"), "icon", user.getString("icon"));
+
         // put the public key credentials parameters
         for (PublicKeyCredential pubKeyCredParam : options.getPubKeyCredParams()) {
           addOpt(
@@ -294,6 +306,7 @@ public class WebAuthnImpl implements WebAuthn {
       WebAuthnCredentials authInfo = (WebAuthnCredentials) credentials;
       // check
       authInfo.checkValid(null);
+
       // The basic data supplied with any kind of validation is:
       //    {
       //      "rawId": "base64url",
@@ -339,6 +352,7 @@ public class WebAuthnImpl implements WebAuthn {
       }
 
       // optional data
+
       if (clientData.containsKey("tokenBinding")) {
         JsonObject tokenBinding = clientData.getJsonObject("tokenBinding");
         if (tokenBinding == null) {
@@ -358,6 +372,7 @@ public class WebAuthnImpl implements WebAuthn {
         }
       }
 
+      final String userId = authInfo.getUserId();
       final String username = authInfo.getUsername();
 
       // Step #4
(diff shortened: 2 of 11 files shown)


vertx-auth-webauthn/src/main/generated/io/vertx/ext/auth/webauthn/AuthenticatorConverter.java
```diff
@@ -60,6 +60,10 @@ public static void fromJson(Iterable<java.util.Map.Entry<String, Object>> json,
             obj.setUserName((String)member.getValue());
           }
           break;
+        case "userId":
```
mnylen, line 63: Realized I modified these by hand... how do I run the code generator?

vertx-auth-webauthn/src/main/java/io/vertx/ext/auth/webauthn/impl/WebAuthnImpl.java
```diff
@@ -159,9 +159,17 @@ public WebAuthn authenticatorUpdater(Function<Authenticator, Future<Void>> updat
...
   @Override
   public Future<JsonObject> createCredentialsOptions(JsonObject user) {
+    String userId;
+    if (user.getString("id") != null) {
+      userId = base64UrlEncode(user.getString("id").getBytes());
+    } else if (user.getString("rawId") != null) {
+      userId = user.getString("rawId");
+    } else {
+      userId = uUIDtoBase64Url(UUID.randomUUID());
+    }
```
mnylen, line 169: Backwards-compatibility wise I'm not sure about this.

Previously the "id" attribute did not end up in the generated JSON at all, but the documentation here https://vertx.io/docs/vertx-auth-webauthn/java/#_create_a_registration_request said it should be a base64url string. The documentation also instructed adding "rawId".

Because of that, it is likely that there are users already out there who have set the "id" and "rawId" attributes who either didn't realize that those were not properly reflected in the generated JSON, or noticed it but didn't care.

So if we now start actually using the "id" attribute (or "rawId"), that might come as surprise to the users when upgrading. Especially if they were already passing the "id" as base64url encoded value, it would now get double-encoded.

Any ideas how to handle this? I see the options are:

1. Make this a potentially breaking change (can this be done in 4.3 -> 4.4?)

2. Rename "id" as "handle" -> no backwards compatibility issues:
   * Any users that were passing "id" or "rawId" would not see any effect, as those were unused previously, and the fallback to base64url encoded UUID would still be in place.

vertx-auth-webauthn/src/test/java/io/vertx/ext/auth/webauthn/DummyStore.java
```diff
@@ -20,17 +21,25 @@ public void clear() {
   }
 
   public Future<List<Authenticator>> fetch(Authenticator query) {
+    if (query.getUserName() == null && query.getCredID() == null && query.getUserId() == null) {
```
mnylen, line 24: I thought it was better to explicitly fail the query in the tests if all conditions were null (which should not happen with proper usage / implementation)

vertx-auth-webauthn/src/test/java/io/vertx/ext/auth/webauthn/WebAuthNTest.java
```diff
@@ -43,27 +46,61 @@ public void testFIDORegister(TestContext should) {
       .authenticatorFetcher(database::fetch)
       .authenticatorUpdater(database::store);
 
-    database.add(
```
mnylen, line 46: Instead of starting the register test with the authenticator already in database, I changed the test to test that the authenticator was persisted in the database _after_ registration.

vertx-auth-webauthn/src/test/java/io/vertx/ext/auth/webauthn/WebAuthNTest.java
```diff
@@ -89,6 +129,32 @@ public void testFIDOLogin(TestContext should) {
...
+
+        final User user = fn.result();
+        should.assertEquals(userId, user.principal().getString("userId"));
+
+        test.complete();
+      });
+  }
+
+  @Test(timeout = 1000)
+  public void testFIDOLoginWhenNoAuthenticatorsFoundByCredID(TestContext should) {
```
mnylen, line 141: Just a little sanity check test to make sure that authenticate() fails if no authenticators are found by the credential id.

mnylen:
~For some reason the eclipsefdn/eca action doesn't find my ECA, which I have signed. The Eclipse Foundation portal shows it as signed, the email address should match the email address on the commit, and the GitHub username should match the username configured on the Eclipse Foundation account. :/~

Seemed to have resolved itself.

mnylen:
Updated the documentation also. Added a mention that it's up to the caller of `createCredentialsOptions` and `getCredentialsOptions` to store the challenge somewhere (with some ideas for where), so it can be retrieved in the second step.

Have to say, I don't super love that the `authenticate` method is the entrypoint for both registration and login. If the implementer isn't careful, it's all too easy to accept [AuthenticatorAttestationResponse](https://developer.mozilla.org/en-US/docs/Web/API/AuthenticatorAttestationResponse) (webauthn.create) where [AuthenticatorAssertionResponse](https://developer.mozilla.org/en-US/docs/Web/API/AuthenticatorAssertionResponse) (webauthn.get) is expected. This could open up a door to register an authenticator to another account through login.

To tackle this somehow, mentioned in the documentation that, when storing the challenge, implementer should scope the stored challenge to either login or registration, so it can only be used in that context. But maybe this is something we should tackle at the API level to prevent possibility for the user to make such mistake.

Issue #583, opened 2022-08-14
Invalid JWT: Token must be a short-lived token (60 minutes) and in a reasonable timeframe. Check your iat and exp values in the JWT claim


afloarea:
### Version

4.3.3

### Context

I was trying to connect to google spreadsheets API using a service account. After some trial and error, I figured I can use something of the sorts:
```
final var auth = GoogleAuth.create(vertx, serviceAccountConfig); // service account config file downloaded from google
final var client = WebClient.create(vertx);

final var oath2Client = OAuth2WebClient.create(client, auth)
	.withCredentials(new Oauth2Credentials()
		.setFlow(OAuth2FlowType.AUTH_JWT)
		.setJwt(new JsonObject().put("scope", String.join(" ", List.of("https://www.googleapis.com/auth/spreadsheets","https://www.googleapis.com/auth/drive")))));
```

At first, it seems to work and it looks pretty cool and simple, however, after about an hour I start to get an error:
`Invalid JWT: Token must be a short-lived token (60 minutes) and in a reasonable timeframe. Check your iat and exp values in the JWT claim after one hour`

Trying to investigate, the internet says that this is probably caused by the clock being out of sync (NTP). However this is was not the case. After more digging I found out that after expiration, the same flow (JWT_AUTH) was being performed again but it failed because of an invalid `iat` field. It turns out that the JsonObject passed with `.setJwt(...)` was altered and new fields were added by the JWT::sign(JsonObject,JWTOptions) method (which set the `iat` field on the first call to google API). After expiration, the sign method runs again, only this time it finds that the payload already has an `iat` field and so, it leaves it as is, which google does not like and returns the before-mentioned error.

### Do you have a reproducer?

Unfortunately I do not, as it involves credentials.

### Steps to reproduce

1. obtain a pair of credentials for service account in the form:
```
{
	"type": "service_account",
	"project_id": "<proj-id>",
	"private_key_id": "<id>",
	"private_key": "<private_key_cert>",
	"client_email": "email@test.com",
	"client_id": "<client_id_value>",
	"auth_uri": "https://accounts.google.com/o/oauth2/auth",
	"token_uri": "https://oauth2.googleapis.com/token",
	"auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
	"client_x509_cert_url": "<cert_url_value>"
}
```

2.  Create the web client as mentioned above
4.  Make a request to some google api
[...]

PR #585, opened 2022-09-04
#583 Do not mutate the jwt object provided by the user

afloarea:
Motivation:

#583  Avoid mutating the JWT object passed by the user and fix the generation of `iat` field which only worked on the first call.

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java
@@ -441,7 +441,7 @@ public class OAuth2AuthProviderImpl implements OAuth2Auth, Closeable {
 
         case AUTH_JWT:
           params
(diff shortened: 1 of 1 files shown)


pendula95:
I would rather have this copy generated [here](https://github.com/vert-x3/vertx-auth/blob/0564883b03b8c4122a11d397d396fe646046e5c1/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java#L444)

It is more clear to have it outside of void method then to modify state of objects inside the void function.

Something similar has been done already [here](https://github.com/vert-x3/vertx-auth/blob/9642af0a60abda2863484369fa911f214fc21bc6/vertx-auth-jwt/src/main/java/io/vertx/ext/auth/jwt/impl/JWTAuthProviderImpl.java#L185) With your modification double copy would be performed.

afloarea:
Thank you for your input, @pendula95 .
I have updated the PR according to your suggestion.

pendula95:
LGTM

pmlopes:
Thank you, @afloarea and @pendula95 !

PR #599, opened 2022-10-26
Fix #589

### LINKED ISSUE #589: DuplicatedContext is replaced by EventLoopContext
### Version

4.3.0 - 4.3.3

### Context

The DuplicateContext that is bound to a HTTP request is replaced by EventLoopContext when calling another service using OAuth2WebClient with client_credentials grant and access_token is expired.

So everything works fine until the token expires, then the request that will trigger re-authentication (request new access_token using client_credentials) will loose it's DuplicatedContext and there for everything that has been put into the ContextData is lost.

The problem we are hitting is exactly here:
https://github.com/vert-x3/vertx-auth/blob/c097a5cad0b4b3eeca88e944e86b5153a124fcf2/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java#L494

If no refresh_token is found the future is failed using an old context that OAuth2AuthProviderImpl keeps as member. Should this not simply use Future.failed(...) ?

### Do you have a reproducer?

https://github.com/runtarinn/duplicate-context-reproducer

### Steps to reproduce

`./gradlew test`

Hope this is enough details provided
/Rúnar

runtarinn:
Fix issue #589 which was introduced as part of the https://github.com/vert-x3/vertx-auth/commit/4064b554aabe7536887fb660c262821c7b90e605

This will most likely not fix all issues introduced in this commit

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2API.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2API.java
@@ -48,12 +48,10 @@ public class OAuth2API {
   private static final Logger LOG = LoggerFactory.getLogger(OAuth2API.class);
   private static final Pattern MAX_AGE = Pattern.compile("max-age=\"?(\\d+)\"?");
 
-  private final ContextInternal ctx;
   private final HttpClient client;
   private final OAuth2Options config;
 
   public OAuth2API(Vertx vertx, OAuth2Options config) {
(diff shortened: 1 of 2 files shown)


pmlopes:
Yes, this is a good start, I'm also addressing the issue for other providers, the "main" problem is that internally we're mixing callbacks and futures, which is not a problem itself, except that some of the callbacks, do require the context to be correct and nothing was being done. Future compositions do address this for us.

pmlopes:
This PR is quite diverged from master and as we moved to Future only APIs I believe this issue isn't relevant anymore.

PR #608, opened 2023-01-20
Security audit implementation. This allows a custom logger to ship security events to a EDR/XDR system OOTB

pmlopes:

Motivation:

Initial prototype for discussion


pmlopes:
@photomorre this is the base for XDR/EDR logging. Once this is merged, the counter part will be merged in web and enabling it is just a matter of adding a top level route handler and configure a java logger.

pmlopes:
For non-disclosure reasons, the log will not dump the user information other than the subject id/name. This prevents unintended leaks of information. This is not supposed to be used for debugging, but for monitoring.

PR #618, opened 2023-02-26
doc: replace mention of deprecated authenticate

milgner:
I was just getting started reading up on `vertx-auth` and found that the first link already sent me to a method that was marked as deprecated. So I replaced the link with the newer version and added a bit of explanation

Also note that the "edit" links on the homepage are still linked to the old Github org and need to be updated to use `eclipse-vertx`. I didn't find whether those are publicly editable, though.

### DIFF
--- a/vertx-auth-common/src/main/asciidoc/index.adoc
+++ b/vertx-auth-common/src/main/asciidoc/index.adoc
@@ -60,19 +60,11 @@ To find out what a particular {@link io.vertx.ext.auth.authorization.Authorizati
 
 == Authentication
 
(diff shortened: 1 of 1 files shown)


milgner:
For some reason the ECA validation doesn't seem to go through. Since I just signed, I'm gonna wait a bit and see whether it works tomorrow or so.

milgner:
ECA green now, gonna have to check why docgen CI fails... :flushed:

pmlopes:
Thanks! I went on a massive review and updated all the deprecated API usage, including examples!

PR #621, opened 2023-03-02
Implement basic support for security policies (which enables ABAC)

pmlopes:
This PR allows defining security policies.

Security policies are a set of rules that when matched against a user object entitle the user a set of permissions.

1. The matching is done by subject id (username/handleId) and/or attributes and/or resources.
2. Attributes are matches against both the user object "principal" and "attributes" fields using a json pointer notation
3. Resources are matches as currently supported by vertx-auth

The PR introduces a AuthZ provider that given a collection of policies, entitles the user to all listed permissions that matches the rules above.

For example, we can entitle all operations to an "admin" user by defining the rule:

```json
{
  "name": "administrator",
  "subjects": [
    "admin"
  ],
  "authorizations": [
    {
      "type": "wildcard",
      "permission": "web:*",
      "resource": "/resources"
    }
  ]
}
```

In this case a match is based on the user id. Extra rules can be defined such as:

```json
{
  "name" : "administrator with DELETE rights",
  "attributes" : {
    "/principal/amr": {
      "eq": "mfa"
    }
  }
  "authorizations" : [ {
    "type" : "wildcard",
    "permission" : "web:DELETE",
    "resource": "/some/special/resource"
  } ]
}
```

In this case, only users that have 2nd factor authentication are allowed to delete the resource `/some/special/resource`

The model attempts to looselly follow the ABAC model but is not intended to be exact as in the definition to avoid changing the internals of vert.x in a system that has been proven to work in the 4.x.x series.


pmlopes:
@stephanebastian @francoisprunier working with the community, we would like to address the concrete support for ABAC authorization. In order to solve the gap and avoid changing the current internals, I'm proposing a simple policy format and authz provider that can assign permissions to a user based on a policy.

Given the flexibility of the current authz, the proposed ABAC support can work together or in isolation with the current RBAC rules. For example, we can have both their permissions read from a database following the RBAC model, plus a set of policies can be used in a policy authz provider and permissions can get assigned at runtime to a user based on their subject name, attributes or resource.

Could you please have a look, if everything is correct in terms of API usage?

francoisprunier:
Hi Paulo,

Generally speaking, looking back at that API, I think what's a bit confusing is that there are authorizations you **have** as a user, and authorizations you **check** (for an API call for example), but they're both called Authorization because they're conceptually the same.

I've had a look at your changes, here are my observations:
- Purely on naming, I'd maybe change AuthorizationPolicyProvider to PolicyBasedAuthorizationProvider
- I'd make Policy inherit from Authorization (call it PolicyBasedAuthorization) to implement the matching logic inside `boolean match(AuthorizationContext context)`
- Then Inside PolicyBasedAuthorizationProvider instantiate a AuthorizationContext(user) and all you have to do (I think) is a loop on policies to call policy.match(authorizationContext)
- Attributes then becomes a simple bean

I hope that makes sense ? Happy to discuss it further if needed

pmlopes:
I don't think we can have `Policy` implement `Authorization` for a couple of reasons:

1. If we did it, then at regular RBAC we could mix Authorizations and policies, and in that case it would be almost impossible to ever match. But this is probably not relevant as it would be a wrong API usage (still confusing though).
2. The matching algorithm handles variables and wildcards from left to right (i.e.:  authzA.verify(authzB)) in a policy wildcards are not allowed as we need to match in reverse.

If you are OK with it, I'll keep `Policy` as a data object independent of `Authorization` which explicitly defines a different use case in terms of API. And enforces the correctness of the checks.

stephanebastian:
Hi Paulo and Francois,

I am not up-to-date on ABAC, so before looking at the PR, I glanced over the following documents:
- https://www.forgerock.com/what-is-rbac-vs-abac
- https://kubernetes.io/docs/reference/access-authn-authz/abac/
- https://docs.aws.amazon.com/IAM/latest/UserGuide/introduction_attribute-based-access-control.html
- https://learn.microsoft.com/en-us/azure/role-based-access-control/conditions-overview

Also, I haven't worked on Vertx Authz API for a long time. So let me first write down a quick refresher (for me ;) of the state of the current Vertx 4.x API, where we've got few moving parts:
- **Authorization**: In theory, it represents any kind of authorization (role, permission, time-based, and, or , not, context-based: ip-adress, etc.)
- **AuthorizationContext**:  holds properties used to match authorizations
- **AuthorizationProvider**: Asynchronously fetch authorizations of the current User

When we refactored the older API, one of the goal was to provide a lean but powerful API that would scale beyond RBAC. It seems to me that we are about to see if it holds its promises ;).

Unless I'm missing something, I believe we can already do some ABAC with the current Vertx API. One of the reason is that Authorization is not limited to role or permission as it is a fairly broad notion. For instance, if you want to give access to some resource from 8am untill 10am to all users holding the permission '007', we could define the following AuthorizationHandler (in Verxt Web):
```
...
  AuthorizationHandler.create(AndAuthorization.create().addAuthorization(TimeBasedAuthorization.create(8am, 10am)).addAuthorization(PermissionBasedAuthorization.create("007"))
...
```
As you know TimeBasedAuthorization is not shipped with Verxt 4 but it's just an example of things doable today with a custom implementation if/as needed. 

To me, it seems that ABAC means different things. For instance:
1. As an admin, I want to protect a specific resource based an some attribute (time, ipaddress, attributes from the user, etc.) and/or permission/role/wildcard. As I said, this is somewhat working today but probably need more testing. For instance I'm not sure we could only grant access to users that have 2nd factor authentication but I may be wrong.
2. As an admin, I want to configure an AuthorizationProvider that will give a bunch of authorizations to all users matching the conditions I specify (2nd form auth, time, ipaddress, attributes from the user, etc.)

As far as I can tell the goal of this PR is 2) right?
If that is the case, this is great but I would suggest to provide full support for 1) as well
Also I am wondering if the ABAC provider belongs to vertx-auth-common of should live in its own sub-project ?

I am also available to discuss it further if needed
Cheers

pmlopes:
I like the idea of splitting to an ABAC module for a complete implementation if needed. Indeed, this PR was only proposing the basic functionality.

Issue #541, opened 2022-03-09
Random errors and 403 when clearing authorizations with concurrent requests


bfreuden:
### Questions

Related question: https://groups.google.com/g/vertx/c/vTCLEQsv1N0

### Version

4.2.5

### Context

I have a Vert.x 4 backend managing authorizations that vary during the lifetime of users (they can have new permissions, new roles, then those roles are revoked... stuff like that).
Because my backend is likely to revoke permissions or roles, when it is the case, I need to clear user permissions and fetch them again from the database. So I'm doing something like this:

```
user = routingContext.user()
user.authorizations().clear(mongoAuthorization.getId())
mongoAuthorization.getAuthorizations(user)
```

The problem is: when my UI is issuing parallel requests, sometimes the user has no permission at all, sometimes I get ConcurrentModificationExceptions.

### Do you have a reproducer?

Here it is (no Mongo inside, just plain PropertyFileAuthentication):
[https://github.com/bfreuden/vertx-reproducers/tree/master/concurrent-authz-issue](https://github.com/bfreuden/vertx-reproducers/tree/master/concurrent-authz-issue)

### Steps to reproduce

Simply run the unique unit test of the reproducer project multiple times until you get an error (happens most of the time).

pendula95:
The issue here is that you are iterating over the Collection while it is being cleared by other thread. The Collection in background should be made thread safe so this can not happen.

bfreuden:
@pendula95 thank you so much for your comment!

That's not the only issue unfortunately.
You can comment out the iteration over the authorizations (line 115 to 123) if you wish...:

```
                    // try to match a permission: this can throw an error sometimes (see out.txt stack traces)
                    if (PermissionBasedAuthorization.create("do:stuff").match(user)) {
//                        Authorizations authorizations = user.authorizations();
//                        // this iteration is also likely to raise ConcurrentModificationException on authorizations
//                        for (String providerId : authorizations.getProviderIds()) {
//                            Set<Authorization> authorizationSet = authorizations.get(providerId);
//                            StringBuilder builder = new StringBuilder();
//                            for (Authorization authorization : authorizationSet) {
//                                builder.append(authorization.toString());
//                            }
//                        }
                        replySuccess(routingContext, "<h1>protected page</h1>");
                    } else {
                        replyNotAuthorized(routingContext, "<h1>not authorized</h1>");
                    }

```

... but you still get 403 errors and and ConcurrentModificationException like this...

```
[...]

PR #635, opened 2023-03-17
Initial work, Authorizations are now CoW Immutable which should solve…

pmlopes:
… issues with multiple access across event loops


pmlopes:
@bfreuden this should address the issue #541 

The idea is that in vert.x 5 as we can break APIs we make the Authorizations a Immutable data structure and apply CoW semantics so multiple threads can access and see the data they saw when started, and not be afected by future mutations.

pmlopes:
The PR is still draft as I'd like to get rid of the `add` method which causes too much copies (ATM it only affects 20 places in tests) and your reproducer is always green for me.

pmlopes:
The reproducer now works 100% good and it takes 1.3s to execute vs 2.6s with lock

bfreuden:
@pmlopes 
Wow! Thank you so much! :+1: 
Can't wait to try it out!

Issue #659, opened 2023-08-29
OAuth2AuthProviderImpl loses Access Token after introspect call


chrispatmore:
### Version

Version: 4.3.8 +

### Context

Whilst investigating [654](https://github.com/eclipse-vertx/vertx-auth/issues/654), I discovered there is an issue when the AT does not validate as a JWT. In these cases (either its an invalid JWT as the code is concerned e.g. [654](https://github.com/eclipse-vertx/vertx-auth/issues/654). or it is just an opaque token) the code then attempts a token introspect [here](https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java#L303) if this passes we know the token is valid and the `User` is created [here](https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java#L328). However at this point the `access_token` is lost, the user placed into the context has no `access_token` so when I then try to make a `userInfo` call or some other request to the provider, I get rejected as I supplied an invalid token (no token, for example [here](https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java#L474)). I have also noticed that since the version I am testing on a `userInfo` introspect has been added if there is no introspect endpoint [here](https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java#L269) and this adds the token back to the user [here](https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java#L280)

In short, the `OAuth2AuthProviderImpl` should be setting the `access_token` back into the `User` object when the token is valid, like happens when the token is validated using the userInfo endpoint. So that the token can then be used for future actions 

**Edit:** This also then sets the `subject` in the user wrong in some cases, for example in the keycloak case, the json returned back from the introspect endpoint contains `username` which is then used to set the subject instead of the `sub` field in the response

### Do you have a reproducer?

No

### Steps to reproduce

[...]

PR #660, opened 2023-08-30
fix: ensure access_token is available after introspect

chrispatmore:
Update the code to ensure the access token is available after validating the token using token introspect

Motivation:

Contributes to: https://github.com/eclipse-vertx/vertx-auth/issues/659


chrispatmore:
I would prefer not to put the token introspect response JSON as the principal JSON as I don't think they're really the same thing, and I have seen (Keycloak for example) returns `username` as well as `sub` and then when that json is placed as the principal (not under idToken or other) the User created returns the `username` as the subject() instead of the `sub` which is what is returned if the token validates / if there is an `idToken`

But. I can see that being a breaking change for those relying on it behaving as is at the moment, so I could accept that I just have to include a workaround to fix the subject post authentication

pmlopes:
We can accept this behavior change as we are targetting to 5.0.0

PR #661, opened 2023-09-12
Added support for revoking KeyCloak tokens

Ernstsen:
## Changes
The revocation endpoint was forced to be null, when using the keycloak oauth2 provider. 

As per the keycloak documentation, there is an endpoint for revoking tokens: https://www.keycloak.org/docs/latest/securing_apps/#_token_revocation_endpoint

This would enable logging users out with API instead of redirects with keycloak.

## Motivation:

I need to be able to log out users, when using the keycloak OAuth2 provider.

This is currently not possible as the ``revoke`` method requires the ``revocationUrl != null``  and the ``logout`` method in ``OAuth2API`` is not accessible from outside the ``OAuth2Auth`` object.

This simple change allows users to use the flow described in the current vertx.io documentation for token management: https://vertx.io/docs/vertx-auth-oauth2/java/#_token_management

## Conformance:


pmlopes:
This is a great catch, thanks!

Issue #654, opened 2023-06-08
OAuth2 Auth provider incorrectly validating Access Tokens 


chrispatmore:
### Questions

Do not use this issue tracker to ask questions, instead use one of these [channels](https://vertx.io/community/). Questions will likely be closed without notice.

### Version

Version: 4.3.8

### Context

Whilst trying to set up an OIDC client with discovery against a keycloak instance, I discovered that the access token which is in JWT form was not being decoded and added to the session as I expected it would be. The non decoded format is present on the session. 

### Steps to reproduce

1. Run a keycloak instance somewhere and create a realm and client and a user
2. Set up the vertex OIDC client with discovery
3. Have an endpoint that prints out the session
4. Run the program and log in and inspect the session

### Extra

The reason the decoded form is not stored is because of https://github.com/eclipse-vertx/vertx-auth/blob/f0215a105380d47ecdefa260e2ebe0aa321a5297/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java#L542
Which calls `validToken` Which internally does https://github.com/eclipse-vertx/vertx-auth/blob/f0215a105380d47ecdefa260e2ebe0aa321a5297/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java#L643

With the comment
```
// https://openid.net/specs/openid-connect-core-1_0.html#  $3.1.3.7.
// The Client MUST validate that the aud (audience) Claim contains its client_id value registered at the Issuer
// identified by the iss (issuer) Claim as an audience. The aud (audience) Claim MAY contain an array with more
// than one element. The ID Token MUST be rejected if the ID Token does not list the Client as a valid audience,
// or if it contains additional audiences not trusted by the Client.
```

This fails for the keycloak access token as the `aud` is not the client ID it is `account`

`https://openid.net/specs/openid-connect-core-1_0.html#  $3.1.3.7` refers to the section for `ID Token Validation`. In this case the token is not an ID token so should not be subject to the same validation logic. The validation logic for access tokens is stated in section `3.2.2.9`

Therefore there are a couple of options / parts:
[...]

PR #673, opened 2024-02-08
OAuth2 Auth provider incorrectly validating Access Tokens

byn-florian-posch:
Motivation:

PR containing code for fix in #654

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java
@@ -32,13 +32,12 @@ import io.vertx.ext.auth.authentication.TokenCredentials;
 import io.vertx.ext.auth.authentication.UsernamePasswordCredentials;
 import io.vertx.ext.auth.impl.jose.JWK;
 import io.vertx.ext.auth.impl.jose.JWT;
-import io.vertx.ext.auth.oauth2.OAuth2Auth;
-import io.vertx.ext.auth.oauth2.OAuth2FlowType;
-import io.vertx.ext.auth.oauth2.OAuth2Options;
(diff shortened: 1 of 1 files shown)


vietj:
I can't see any test for this

vietj:
@pmlopes could you have a look ?

byn-florian-posch:
> I can't see any test for this

I didn't write a test because I couldn't see any for the provider. A full maven build for the whole vertx-auth module passes locally but that's as far as I ran tests locally.

byn-florian-posch:
Test issues partially fixed, reusing the _audience_ ArrayList from JWTOptions did unintendedly modify its content, using a copy now.

byn-florian-posch:
Now only the following two tests are failing in _OAuth2Keycloak14IT_ 

- discoverGetTokenFromFrontEndPerformAuthWithBackend
- discoverGetTokenFromFrontEndPerformAuthWithBorkendWillFail

Tests are expecting validation to fail but that isn't happening with the suggested ID-token-only validation. I don't want to tamper with existing tests too much, so your opinion on this one would be appreciated.

byn-florian-posch:
@pmlopes @vietj May I ask for an update?

My recent changes to suggested fix from https://github.com/eclipse-vertx/vertx-auth/issues/654 did resolve some of the regression but the ones are broken with new logic only validating ID-tokens. Writing a meaningful test is above my paygrade and I wouldn't dare to just remove the test cases. 

Let me know what you think,

byn-florian-posch:
There hasn't been any reply, neither on the related [issue](https://github.com/eclipse-vertx/vertx-auth/issues/654) nor here for months and I'm not sure how else to address this.

byn-florian-posch:
We decided to tackle this issue by this issue by managing audiences in our application in a controlled manner, so no need to fix token handling from our perspective -> closing the PR and leave it to the creator of the related [issue ](https://github.com/eclipse-vertx/vertx-auth/issues/654)how to deal with it. 

See also https://github.com/eclipse-vertx/vertx-auth/issues/654

PR #704, opened 2024-11-25
Updated docs for webauth/webauthn4j about storage

FroMage:
We want to make sure users implement credential storage properly, so we tell them what to do and what to watch out for.

This includes warning them about unicity of credential ID and user names, multiple credentials and what to do when trying to store multiple credentials for the same user name.


vietj:
This looks great @FroMage

ynojima (approved):
Thank you for update! my concerns are all resolved.

PR #705, opened 2024-12-04
Rename Authenticator#userName to Authenticator#username

vietj:
Motivation:

The `Authenticator#userName` field name is used by the construction of the User instance based on the data object to JSON converter. The `User#subject()` method maps the `username` value of the principal built from the JSON representation, leading to return `null` instead of the user name.

Change:

Rename `Authenticator#userName` as well as the property accessors to username and update the tests accordingly.


FroMage (approved):
Great!

PR #707, opened 2024-12-18
io.vertx.ext.auth.authentication.Credentials should be a data object interface

vietj:
Motivation:

`io.vertx.ext.auth.authentication.Credentials` is a code generated object, however its implementations are data objects. Generated code declaring `Credentials` as argument or return type does not accept `Credentials` implementation since the type is translated to the shim equivalent.

Changes:

Make `Credentials` a data object.


vietj:
See https://github.com/vert-x3/vertx-rx/issues/321

PR #714, opened 2025-03-09
Included client_assertion_type and client_assertion when other auth m…

### LINKED ISSUE #713: Allow specifying clientId together with clientAssertion/clientAssertionType
Currently it's only possible to either specify the `clientId` or use `clientAssertion`/`clientAssertionType`.

However some provider require the use of `clientId` together with `clientAssertion`/`clientAssertionType`
e.g. Microsoft requires a `clientId` if the certificate credential is used see: https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#request-an-access-token-with-a-certificate-credential
and I think also keycloak requires clientId always to be present:
https://github.com/keycloak/keycloak/blob/5387aef0fa727ea5cae4816f682ec72798fabaa4/services/src/main/java/org/keycloak/protocol/oidc/endpoints/request/AuthorizationEndpointRequestParserProcessor.java#L119-L123

The code responsible for it is in:
https://github.com/eclipse-vertx/vertx-auth/blob/be9744623224e45eb8df868fa3ff23d193e98699/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2API.java#L184-L194

I guess this could be relaxed to only check if clientSecret is null and allow both `clientId` and `clientAssertion`/`clientAssertionType`.

Happy to provide a PR for this.

pendula95:
Fixes #713

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2Options.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2Options.java
@@ -296,6 +296,7 @@ public class OAuth2Options {
 
   public OAuth2Options setClientAssertionType(String clientAssertionType) {
     this.clientAssertionType = clientAssertionType;
+    this.useBasicAuthorization = false;
     return this;
   }
 
@@ -623,28 +624,18 @@ public class OAuth2Options {
         case AUTH_CODE:
         case AUTH_JWT:
         case AAD_OBO:
-          if (clientAssertion == null && clientAssertionType == null) {
+          if (clientAssertionType == null) {
             // not using client assertions
             if (clientId == null) {
               throw new IllegalStateException("Configuration missing. You need to specify [clientId]");
             }
-          } else {
-            if (clientAssertion == null || clientAssertionType == null) {
-              throw new IllegalStateException(
-                "Configuration missing. You need to specify [clientAssertion] AND [clientAssertionType]");
-            }
           }
           break;
         case PASSWORD:
-          if (clientAssertion == null && clientAssertionType == null) {
+          if (clientAssertionType == null) {
             // not using client assertions
             if (clientId == null) {
               LOG.debug("If you are using Client Oauth2 Resource Owner flow. You need to specify [clientId]");
-            }
-          } else {
-            if (clientAssertion == null || clientAssertionType == null) {
-              throw new IllegalStateException(
-                "Configuration missing. You need to specify [clientAssertion] AND [clientAssertionType]");
             }
           }
           break;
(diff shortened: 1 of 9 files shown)


tsegismont:
@pmlopes any chance for you to take a look?

tsegismont (commented):
Thank you @pendula95 

Is there any possibility to add a test with Keycloak?

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2Options.java
```diff
@@ -296,6 +296,7 @@ public String getClientAssertionType() {
 
   public OAuth2Options setClientAssertionType(String clientAssertionType) {
     this.clientAssertionType = clientAssertionType;
+    this.useBasicAuthorization = false;
```
pendula95, line 299: This is a shortcut that might introduce unexpected behaviors for some users

pendula95 (commented):
I am 100% confident about this change. Looks like client assertion as client authentication method was missed and previous code would not pass needed tests.
I have tried to work around current code to not introduce breaking changes or new api features but as result there are some maybe unexpected behaviors. It might be helpful to have  discussion about this changes.

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/OAuth2Options.java
tsegismont: I'm sorry but I'm not an expert in that area. Can you elaborate about why this might come unexpected for some users? Especially as it seems you say the current behavior is broken in the other comment

pendula95: There are 2 ways how to use assertion. Both defined: https://datatracker.ietf.org/doc/html/rfc7521
* [Using Assertions as Authorization Grants](https://datatracker.ietf.org/doc/html/rfc7521#section-4.1)
* [Using Assertions for Client Authentication](https://datatracker.ietf.org/doc/html/rfc7521#section-4.2)

This PR fixes incorrect implementation for Using Assertions for Client Authentication

As test were missing that would catch this, it was overlooked. This has been fixed by the PR and I am quite familiar with this.
On the other hand most of the code that is present its goal was to facilitate Using Assertions as Authorization Grants which I have limited knowledge. And looking at the test non of them  validate the functionality so I am not 100% sure that something was not broken by this PR.

pmlopes (approved):
I think this PR is correct and makes sense. Indeed as @pendula95 noticed, the assertion implementation was not considering client authentication. The original code was mainly focused on the authorization grants, so we could support service accounts (how Google like to call it) and On-Behalf-Of (like how Azure likes to call it).

If I remember correctly the extra null checks where there to ensure a smooth interop with Vert.x Web. If this PR passes the existing + new tests and does not introduce any side-effects/regressions in Vertx-Web, it gets my approval 👍

tsegismont:
Hi @pendula95,

We've merged #753 as a minimal fix for #713, that allows client_id to coexist with client_assertion/client_assertion_type in token requests.

Your PR goes further, it would be great to have it. Would you mind updating it?

PR #715, opened 2025-03-27
Scope from jwt should be correctly propagated to the user principal

pendula95:
Motivation:

Scopes that are present inside a JWT MUST be propagated to the user, specifically to the principal. 

I caught this bug while working on OpenAPI specification where I used RouterBuilder with automatic security handler that was also generated from spec. If request landed with a bearer token (bearer only handler) OAuth2AuthHandler failed to validate the scopes despite having set scopes on the OAuth2AuthHandler. Handler allowed request with JWT that had invalid scopes. This is because handler only checks if user has scopes, if scopes are not present validation is not done.

 https://github.com/vert-x3/vertx-web/blob/a7917a02e0d6ff6b498b17a9ad74094f564b4ae7/vertx-web/src/main/java/io/vertx/ext/web/handler/impl/OAuth2AuthHandlerImpl.java#L347-L380

I also noticed that after my changes some tests failed. After taking a closer look and double checking the tests I positives there was a bug in tests also hence I updated them also.

I would like also to backport this to vertx 4 as it is a critical bug.


vietj:
@pmlopes can you have a look please ?

pendula95:
@pmlopes a gentile ping for this

PR #719, opened 2025-04-29
Update MetaDataServiceTest.java

pmlopes:
Disable old MDS2 test as it is long deprecated and certificates expired


tsegismont:
Thanks @pmlopes 

As we are moving to Central for snapshot and releases management, we are experiencing issues with builds. So I'll wait for these to be solved be fore merging

tsegismont:
It seems we have an unrelated problem with the container image for Keycloak (outdated)

vietj:
I think this does not target the correct branch

vietj:
I'll try cherry-pick to 4.x and 5

tsegismont:
Fixed by afe9164

PR #723, opened 2025-05-20
Update the routes for linkedin

pmlopes:
Motivation:

The old config was out of date and referred to the v1 api which is gone.

The PR uses the new v2 api and enables auto discovery to enhance security with the use of JWKs


tsegismont:
Can you check the failure please @pmlopes ?

PR #730, opened 2025-09-04
OpenID Connect: Respect the configured flow types if the server sends none

### LINKED ISSUE #729: OpenIDConnectAuth.discover: Only reset supported grant types when the server sends it
### Version

5.0.4

### Context

While using `OpenIDConnectAuth.discover`, I am setting `supportedGrantTypes` to be only `implicit`. In our specific scenario, we are using the OAuth2 handler only to validate JWT access tokens sent by clients, therefore we don't need any other grant type.

After discovery, the `OpenIDConnectAuth` resets `supportedGrantTypes` to `null`. When validating the configuration, `OAuth2Auth` falls back to the default, which is `auth_code, implicit`, leading to the following exception:

```
java.lang.IllegalStateException: Configuration missing. You need to specify [clientId]
	at io.vertx.ext.auth.oauth2.OAuth2Options.validate(OAuth2Options.java:629) ~[vertx-auth-oauth2-5.0.4.jar:5.0.4]
	at io.vertx.ext.auth.oauth2.impl.OAuth2AuthProviderImpl.<init>(OAuth2AuthProviderImpl.java:70) ~[vertx-auth-oauth2-5.0.4.jar:5.0.4]
	at io.vertx.ext.auth.oauth2.OAuth2Auth.create(OAuth2Auth.java:56) ~[vertx-auth-oauth2-5.0.4.jar:5.0.4]
	at io.vertx.ext.auth.oauth2.providers.OpenIDConnectAuth.lambda$discover$1(OpenIDConnectAuth.java:149) ~[vertx-auth-oauth2-5.0.4.jar:5.0.4]
	at io.vertx.core.impl.future.Composition.complete(Composition.java:40) ~[vertx-core-5.0.4.jar:5.0.4]
	at io.vertx.core.impl.future.FutureBase.emitResult(FutureBase.java:68) ~[vertx-core-5.0.4.jar:5.0.4]
	at io.vertx.core.impl.future.FutureImpl.completeInternal(FutureImpl.java:163) ~[vertx-core-5.0.4.jar:5.0.4]
	at io.vertx.core.impl.future.FutureBase.emitResult(FutureBase.java:68) ~[vertx-core-5.0.4.jar:5.0.4]
```

although no clientId should be required, because it was never intended to be used.

A potential solution would be to reset the `supportedGrantTypes` only when the authorization server defines `grant_types_supported` the field in its response. Otherwise the configured grant types should be kept.

### Steps to reproduce

_No response_

### Do you have a reproducer?

_No response_

Traderjoe95:
Motivation:

The `OpenIDConnectAuth` unconditionally resets the `supportedGrantTypes` of `OAuth2Options` to `null`. Therefore, the previously configured grant types are ignored and replaced by the default (`implicit, auth_code`). This triggers a configuration validation exception if no `clientId` is configured. In use cases that only need OAuth token validation functionality, this behavior is undesired, as they might never need a client ID.

This is fixed by only resetting `supportedGrantTypes` when the authentication server sends `grant_types_supported` on its own.

Fixes #729 

Conformance:

### DIFF
--- a/vertx-auth-oauth2/pom.xml
+++ b/vertx-auth-oauth2/pom.xml
@@ -32,6 +32,18 @@
     <vertx.surefire.useModulePath>false</vertx.surefire.useModulePath>
   </properties>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.testcontainers</groupId>
+        <artifactId>testcontainers-bom</artifactId>
+        <version>1.21.3</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
   <dependencies>
     <dependency>
       <groupId>io.vertx</groupId>
@@ -50,8 +62,17 @@
     <dependency>
       <groupId>org.testcontainers</groupId>
       <artifactId>testcontainers</artifactId>
-      <version>1.18.0</version>
       <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>mockserver</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mock-server</groupId>
+      <artifactId>mockserver-client-java</artifactId>
+      <version>5.15.0</version>
     </dependency>
   </dependencies>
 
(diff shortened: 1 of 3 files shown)


vietj:
@Traderjoe95 can you add a test for this?

Traderjoe95:
@vietj sure, I'm working on it. It's been a bit of a hassle, but I'm getting to it.

There's one thing that just came to my mind, and I would like to get your opinion on this: When the user configures a set of supported flow types, and the server also sends a set, wouldn't it be better to use the intersection of both instead of overriding the previous configuration? 

In this scenario, setting the `supportedGrantTypes` means "My application supports the following grant types". The server may specify which grant types it supports, so the resulting set of grant types agreed on by both parties would be the intersection. If it is empty, that would be an error.

If the user or server doesn't set any supported grant types, that means "I support all grant types", if both don't set anything, we fall back to the default.

What do you think? In my opinion that would allow more flexibility than just having the server override everything

Traderjoe95:
I will first push my tests for the original implementation, then in a second commit, I will update to my suggestion so it can be easily reverted if you don't like it

Traderjoe95:
I updated the code to the alternative solution I proposed earlier. Let me know what you think!

PR #732, opened 2025-09-04
OAuth2: Allow empty JWTOptions.audience

### LINKED ISSUE #731: Setting `JWTOptions.audience` to an empty list rejects all tokens
### Version

5.0.4

### Context

In Vert.x 4.x, it was possible to set the required audience for an OAuth2 provider to an empty list, thus effectively disabling the audience check.

This no longer works in Vert.x 5.0, because audiences are compared using `Collections.disjoint(...)`, which returns `true` if any of the parameters is empty. Effectively, a OAuth2 handler configured with `JWTOptions.audience = []` will reject every access token that contains an audience (virtually every token). Here's the snippet for reference:

**Vert.x 4.x**
```java
// OAuth2AuthProviderImpl.java lines 681-699
    if (target != null && target.size() > 0) {
      if (idToken || jwtOptions.getAudience() == null) {
        // https://openid.net/specs/openid-connect-core-1_0.html#  $3.1.3.7.
        // The Client MUST validate that the aud (audience) Claim contains its client_id value registered at the Issuer
        // identified by the iss (issuer) Claim as an audience. The aud (audience) Claim MAY contain an array with more
        // than one element. The ID Token MUST be rejected if the ID Token does not list the Client as a valid audience,
        // or if it contains additional audiences not trusted by the Client.
        if (!target.contains(config.getClientId())) {
          throw new IllegalStateException("Invalid JWT audience. expected: " + config.getClientId());
        }
      } else {
        final List<String> aud = jwtOptions.getAudience();
        for (String el : aud) {
          if (!target.contains(el)) {
            throw new IllegalStateException("Invalid JWT audience. expected: " + el);
          }
        }
      }
    }
```

**Vert.x 5.0**
```java
// OAuth2AuthProviderImpl.java lines 647-662
    if (target != null && target.size() > 0) {
      if (idToken || jwtOptions.getAudience() == null) {
        // https://openid.net/specs/openid-connect-core-1_0.html#  $3.1.3.7.
        // The Client MUST validate that the aud (audience) Claim contains its client_id value registered at the Issuer
        // identified by the iss (issuer) Claim as an audience. The aud (audience) Claim MAY contain an array with more
        // than one element. The ID Token MUST be rejected if the ID Token does not list the Client as a valid audience,
        // or if it contains additional audiences not trusted by the Client.
        if (!target.contains(config.getClientId())) {
          throw new IllegalStateException("Invalid JWT audience. expected: " + config.getClientId());
        }
      } else {
        if (Collections.disjoint(jwtOptions.getAudience(), target.getList())) {
          throw new IllegalStateException("Invalid JWT audience. expected: " + Json.encode(jwtOptions.getAudience()));
        }
      }
    }
```

I think the intention of the `Collection.disjoint(...)` test isn't to reject tokens when the required audiences are configured as empty, so maybe it would be best to do something like this instead:

```java
if (!jwtOptions.getAudience().isEmpty() && Collections.disjoint(jwtOptions.getAudience(), target.getList()))
```

### Steps to reproduce

_No response_

### Do you have a reproducer?

_No response_

Traderjoe95:
Motivation:

Vert.x 5.0 breaks the option to put an empty array for `JWTOptions.audience` in order to disable the audience check. This is fixed by adding a condition to the audience check. Regression tests are provided for this case and also multiple other cases that were missing.

Fixes #731 

Conformance:

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/impl/OAuth2AuthProviderImpl.java
@@ -644,7 +644,7 @@ public class OAuth2AuthProviderImpl implements OAuth2Auth, Closeable {
       }
     }
 
-    if (target != null && target.size() > 0) {
+    if (target != null && !target.isEmpty()) {
       if (idToken || jwtOptions.getAudience() == null) {
         // https://openid.net/specs/openid-connect-core-1_0.html#  $3.1.3.7.
         // The Client MUST validate that the aud (audience) Claim contains its client_id value registered at the Issuer
@@ -654,7 +654,7 @@ public class OAuth2AuthProviderImpl implements OAuth2Auth, Closeable {
         if (!target.contains(config.getClientId())) {
           throw new IllegalStateException("Invalid JWT audience. expected: " + config.getClientId());
         }
-      } else {
+      } else if (!jwtOptions.getAudience().isEmpty()) {
         if (Collections.disjoint(jwtOptions.getAudience(), target.getList())) {
           throw new IllegalStateException("Invalid JWT audience. expected: " + Json.encode(jwtOptions.getAudience()));
         }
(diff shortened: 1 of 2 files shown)


tsegismont (commented):
Thanks for PR @Traderjoe95 

Can you amend the first commit so that we can see it's only a renamed class?

Traderjoe95:
@tsegismont, I simply changed the name back and squashed the commits, that was the quickest. I couldn't come up with a better idea

tsegismont (approved):
Thank you @Traderjoe95 , that looks good to me

PR #733, opened 2025-09-25
Refactor JWT signing algorithm creation

vietj:
Motivation:

JWT requires the creation of signing algorithm to use for the creation and validation of tokens.

This could be used to implement Quic token validation process and thus we could reuse the implementation engine to implement a Quic token validator.

This refactor also ease the creation usage of thread local to replace synchronised blocks around java cryptographic objet non thread safe usage.

Changes:

Extract the sign/verify part into a small package independent of JWT that can be reused in this module.


vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java
```diff
@@ -449,13 +444,13 @@ public JWK(JsonObject json) {
 
           switch (alg) {
             case "HS256":
-              createOCT("HMacSHA256", json);
+              signingAlgorithm = createOCT("HS256", "HMacSHA256", json);
```
pmlopes: The first argument should be the variable `alg`, to simplify any future refactoring

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java
```diff
@@ -449,13 +444,13 @@ public JWK(JsonObject json) {
...
               break;
             case "HS384":
-              createOCT("HMacSHA384", json);
+              signingAlgorithm = createOCT("HS384", "HMacSHA384", json);
```
pmlopes: The first argument should be the variable `alg`, to simplify any future refactoring

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java
```diff
@@ -449,13 +444,13 @@ public JWK(JsonObject json) {
...
               break;
             case "HS512":
-              createOCT("HMacSHA512", json);
+              signingAlgorithm = createOCT("HS512", "HMacSHA512", json);
```
pmlopes: The first argument should be the variable `alg`, to simplify any future refactoring

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java
```diff
@@ -553,7 +563,8 @@ private void createOKP(JsonObject json) throws NoSuchAlgorithmException, Invalid
           .appendBytes(key)
           .getBytes());
 
-      publicKey = KeyFactory.getInstance("EdDSA").generatePublic(new X509EncodedKeySpec(spki));
+      PublicKey publicKey = KeyFactory.getInstance("EdDSA").generatePublic(new X509EncodedKeySpec(spki));
+      return new PubKeySigningAlgorithm(kty, alg, null, publicKey);
```
pmlopes: I think this is inconsistent. If should be like the previous cases, at the end if either public key or private key are not null then create `PubKeySigningAlgorithm` as the JWK may include both

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java
```diff
@@ -588,17 +599,21 @@ private void createOKP(JsonObject json) throws NoSuchAlgorithmException, Invalid
           .getBytes()
       );
 
-      privateKey = KeyFactory.getInstance("EdDSA").generatePrivate(new PKCS8EncodedKeySpec(pkcs8));
+      PrivateKey privateKey = KeyFactory.getInstance("EdDSA").generatePrivate(new PKCS8EncodedKeySpec(pkcs8));
+      return new PubKeySigningAlgorithm(kty, alg, privateKey, null);
```
pmlopes: See note above.

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWS.java
```diff
@@ -289,8 +226,11 @@ private static int getSignatureLength(String alg, PublicKey publicKey) throws No
         case RS512:
         case PS512:
           return 512;
+        case "EC":
+          // Called by test testGenerateNewTokenES256
```
pmlopes: Shouldn't this be in the same group as `EdDSA`, `ES256` and `ES256K` ? Also, Where is this test?

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/algo/MacSigningAlgorithm.java
```diff
@@ -0,0 +1,67 @@
...
+    return new Signer() {
+      @Override
+      public byte[] sign(byte[] data) {
+        synchronized (MacSigningAlgorithm.this) {
+          return mac.doFinal(data);
+        }
+      }
+
+      @Override
+      public synchronized boolean verify(byte[] expected, byte[] payload) {
```
pmlopes: do we need the double synchronized? (not sure)

pmlopes (changes requested):
I added a few questions and some nitpicking :)

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWS.java
pmlopes: The official list of algorithms is here: https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms could it be a typpo?

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/algo/MacSigningAlgorithm.java
vietj: not for now, I will rewrite this code later to use a Netty FastThreadLocal and avoid synchronization instead

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java
vietj: yes that was actually failing a test JWTTest#testOKP and I got it right now

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWS.java
vietj: This is the test `JWTAuthProviderTest#testGenerateNewTokenES256`, for now I'll leave it as is until we investigate the reason

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/algo/MacSigningAlgorithm.java
vietj: which would address https://github.com/eclipse-vertx/vertx-auth/pull/726 as well

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java
vietj: done

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java
vietj: done

vietj:
@pmlopes please take another look

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWS.java
vietj: it does not seem to fail anymore with the current status (I am thinking it might be a transient failure that went away with the final form of the PR).

vietj:
@pmlopes I think we are done with the refactor and the code is ready to be moved to vertx-core as an internal utility

vietj:
@zyclonite I have a default thread local based version, the reason of not using Netty's fast thread local is because those consume a slot in fast thread local registry and cannot be recycled. Netty fast thread local should be static and here we can only use non static thread local due to the per JWK configuration.

vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWS.java
vietj: actually I found the issue, it does not happen with old JDK (like 11) and the error was a regression in the refactor that has been addressed (in OKP key type).

zyclonite:
sounds reasonable and will work, i think i started using the default for my initial tests as well.

PR #735, opened 2025-10-15
RFC-7591 - Dynamic Client Registration

sanjuthomas:
Please refer to the original [proposal](https://github.com/sanjuthomas/vert.x-rfc7591/blob/main/README.md) to learn more about the Dynamic Client Registration and its needs in Vert.x.

This PR contains changes to implement RFC-7591 for Vert.x OAuth. Just so you know, this implementation has been tested with Keycloak for the Client Authenticated Registration Flow mentioned [here](https://github.com/sanjuthomas/vert.x-rfc7591/blob/main/README.md#client-authenticated-registration-flow).

This implementation supports two optional "manage registration flows" for reading and deleting a client.

I would start with the integration test case DCRKeycloak25_0_0_IT to get a jumpstart with this PR.


sanjuthomas:
@pmlopes  - thank you! I'm looking forward to hearing from others in the community.

pmlopes (approved):

vietj:
@sanjuthomas I'll have a look at the PR soon-ish, the goal is to make it for 5.1

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/dcr/impl/KeycloakClientRegistrationImpl.java
```diff
@@ -0,0 +1,88 @@
...
+        dcrRequest.getClientId()), registrationToken, null)
+        .compose(response -> {
+          if (response.statusCode() != 204) {
+            return Future.failedFuture("Bad Response [" + response.statusCode() + "] " + response.body());
+          }
+          return Future.succeededFuture();
+        });
+  }
+
+  private Future<DCRResponse> constructResponse(SimpleHttpResponse response, int expectedStatusCode) {
```
pendula95, line 75: Error handling should be included at minimum with support for error and error_description as stated by the https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.2

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/DCRRequest.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package io.vertx.ext.auth.oauth2;
+
+import io.vertx.codegen.annotations.DataObject;
+import io.vertx.codegen.json.annotations.JsonGen;
+import io.vertx.core.json.JsonObject;
+
+@DataObject
+@JsonGen(publicConverter = false)
+public class DCRRequest {
```
pendula95, line 19: I would suggest to include all possible values that are listed in https://datatracker.ietf.org/doc/html/rfc7591#section-2
This is a finite list so possibility to have these configured is a must, as with current options no real world usage of this this implementation is possible.

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/ClientRegistrationProvider.java
```diff
@@ -0,0 +1,27 @@
...
+
+import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.core.Future;
+
+/**
+ *
+ */
+@VertxGen
+public interface ClientRegistrationProvider {
+  Future<DCRResponse> create(String clientId);
```
pendula95, line 21: Better provide a way to create a client with client metada data (DCRRequest) then just passing just a client id

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/DCRResponse.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+package io.vertx.ext.auth.oauth2;
+
+import io.vertx.codegen.annotations.DataObject;
+import io.vertx.codegen.json.annotations.JsonGen;
+import io.vertx.core.json.JsonObject;
+
+@DataObject
+@JsonGen(publicConverter = false)
+public class DCRResponse {
```
pendula95, line 19: Possibly provide a structure that will capture all meta data returned by the auth server as in advance all possible fields can not be know but should be parsed by the client.

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/ClientRegistrationProvider.java
```diff
@@ -0,0 +1,27 @@
...
+package io.vertx.ext.auth.oauth2;
+
+import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.core.Future;
+
+/**
+ *
+ */
+@VertxGen
+public interface ClientRegistrationProvider {
```
pendula95, line 20: Missing implementation of https://datatracker.ietf.org/doc/html/rfc7592#section-2.2

vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/DCRRequest.java
sanjuthomas: @pendula95 - Keycloak exposes two endpoints to create clients.

1. /realms/{realm}/clients-registrations/default
2. /realms/{realm}/clients-registrations/openid-connect

The second one accepts the attributes mentioned here -> https://datatracker.ietf.org/doc/html/rfc7591#section-2 with the same names and naming conventions. 

We initially decided to use /realms/{realm}/clients-registrations/default - I think it supports most, if not all, of the attributes mentioned in the spec, but with a different naming convention. I was wondering which one we should shoot for?

pmlopes: I'd rather go for the second one since it aligns with the spec and makes this component generic (not highly coupled to keycloak)

sanjuthomas: There is a discrepancy in how Keycloak handles the registration and get endpoints. The registration endpoint returns results in snake case, but the get endpoint returns them in camel case. This behavior will force us to maintain two POJOS or write a custom JSON parser. 

/realms/{realm}/clients-registrations/openid-connect
/realms/master/clients-registrations/default/{client_id}

@pendula95 - any suggestions?

pendula95: Yes, looks like Keycloak DefaultClientRegistrationProvider camel snake case POJOs and OIDCClientRegistrationProvider uses snake case. 
As I am not aware of the agreements that were made before this PR on which one to implement, I would suggest to implement OIDC one as it is more generic and should be supported by other servers as well. If you want to implement both then I would go for maintaining 2 sets of POJOs.

vietj:
what is the status of this PR ?

vietj:
@pmlopes @sanjuthomas can we merge this contribution ?

Issue #739, opened 2026-01-08
Update azure.json test file


tsegismont:
### Version

All versions

### Context

The azure.json test file contains expired certificate information and fails the build:

```
Error:  Errors: 
Error:    JWKTest.loadAzure:180 » Runtime java.security.cert.CertificateExpiredException: NotAfter: Sat Dec 20 20:50:17 UTC 2025
Error:    JWKLegacyBase64Test>JWKTest.loadAzure:180 » Runtime java.security.cert.CertificateExpiredException: NotAfter: Sat Dec 20 20:50:17 UTC 2025
```

The `JWK#loadAzure` test has been marked as ignored until the file is updated.

tsegismont:
Hi @pmlopes 

Could you help us with this please? Ideally, we'd document how to do this so that anyone can do it easily in the future.

pmlopes:
Hey! This JSON should be in sync with: https://login.microsoftonline.com/common/discovery/keys

This is only an issue in this test because, in the context of OIDC/OAuth2 it get's automatically downloaded at runtime by https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/providers/AzureADAuth.java#L63

tsegismont:
OK, so once we update the file, the failure reported above goes away, but then the test fails when calling `jwt.decode`

pmlopes:
The reason for this test was to cover the old `v1` AzureAD tokens. They had a "special" `nonce` field that is not part of the JOSE spec.

https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-common/src/test/java/io/vertx/tests/JWKTest.java#L187

The issue caused lots of problems in the far past: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/815

And it would either require custom calls to the Azure GraphAPI (totally unrelated to GraphQL BTW) or handle it internally.

This test had a old `v1` token to verify that we could parse it.

While `v1` tokens are rare these days, they are still in use for "work/school" microsoft accounts (for reasons).

The API hasn't changed, it is frozen and v2 is the way to go.

To re-enable this test, one needs to create a business app, which will cost money (not sure if one can do this with some Azure free tier. Since it required "enterprise" AzureAD, probably not).

This feature might just be useless today:

[...]

PR #741, opened 2026-01-13
Updates for Azure AD Graph API retirement

tsegismont:
See #739

https://techcommunity.microsoft.com/blog/microsoft-entra-blog/action-required-azure-ad-graph-api-retirement/4090533

We can get rid of code to handle the "nonce" field that isn't part of the JOSE spec.


tsegismont:
@pmlopes would you mind taking a look ?

tsegismont:
The API has been retired already, still this is technically a breaking API change that must be documente properly.

@vietj @pmlopes how about deprecating the corresponding method in `4.x` and `5.0`? We could also remove the code so that calling the methods have no effect at all. But that is optional, I think.

pmlopes (approved):

tsegismont:
I'll go ahead and merge this one. For Vert.x 4.5 and 5.0, I'll send PRs to deprecate `nonceAlgorithm` in `JWTOptions`

tsegismont:
Thanks @pmlopes !

PR #747, opened 2026-04-21
Ignore EmulatorTest#testMetadata

tsegismont:
Ignoring test for now because we need this update in WebAuthn4J: https://github.com/webauthn4j/webauthn4j/pull/1318

But the fix version (0.31.3) requires Java 17+:
https://github.com/webauthn4j/webauthn4j/pull/1221


tsegismont:
Also, the the 0.31 versions requires Jackson 3: https://github.com/webauthn4j/webauthn4j/pull/1222

PR #752, opened 2026-06-12
WebAuthn: Replace expired Android cert

tsegismont:


tsegismont:
@pmlopes fine with you?

PR #753, opened 2026-08-07
Allow clientId together with clientAssertion/clientAssertionType

### LINKED ISSUE #713: Allow specifying clientId together with clientAssertion/clientAssertionType

jnbdz:
Some providers (e.g. Microsoft Entra, Keycloak) require the `client_id` to be present in the token request even when the client authenticates with an assertion, and the OIDC core spec includes `client_id` in its `private_key_jwt` example as well.

Previously `OAuth2API.token()` only sent `client_assertion`/`client_assertion_type` when no `clientId` was configured. This change sends the assertion parameters whenever the client does not authenticate with a client secret, so `client_id` and the assertion can be used together. Client secret authentication still takes precedence and does not send the assertion.

Added `OAuth2ClientAssertionTest` covering:
- `client_id` + assertion sent together
- assertion-only clients (unchanged behavior)
- client secret clients ignore the assertion (unchanged behavior)

Fixes #713


tsegismont (approved):
Thank you @jnbdz , LGTM

PR #754, opened 2026-08-08
Only reset supported grant types when the discovery document sends them

### LINKED ISSUE #729: OpenIDConnectAuth.discover: Only reset supported grant types when the server sends it

jnbdz:
`OpenIDConnectAuth.discover` unconditionally cleared the configured `supportedGrantTypes` before checking whether the discovery document contains the optional `grant_types_supported` field. When the server omits it, the user configuration was lost, validation fell back to the defaults (`auth_code`, `implicit`), and discovery failed with `Configuration missing. You need to specify [clientId]` — even for setups that never needed a clientId, such as implicit-only clients used purely to validate JWT access tokens.

This change keeps the configured grant types unless the server actually announces its own, in which case they still replace the configuration as before.

Added `OpenIDCDiscoveryGrantTypesTest` (local mock discovery server) covering:
- configured grant types survive when the server omits `grant_types_supported`
- server-announced grant types still replace configured ones
- no configured and no discovered grant types still falls back to the defaults and requires a clientId

Fixes #729

### DIFF
--- a/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/providers/OpenIDConnectAuth.java
+++ b/vertx-auth-oauth2/src/main/java/io/vertx/ext/auth/oauth2/providers/OpenIDConnectAuth.java
@@ -135,11 +135,10 @@ public interface OpenIDConnectAuth {
         }
 
 
-        // reset config
-        config.setSupportedGrantTypes(null);
-
         if (json.containsKey("grant_types_supported")) {
-          // optional config
+          // optional config, only replace the configured grant types when the server
+          // announces the ones it supports
+          config.setSupportedGrantTypes(null);
           JsonArray flows = json.getJsonArray("grant_types_supported");
           flows.forEach(el -> config.addSupportedGrantType((String) el));
         }
--- a/vertx-auth-oauth2/src/test/java/io/vertx/tests/OpenIDCDiscoveryGrantTypesTest.java
+++ b/vertx-auth-oauth2/src/test/java/io/vertx/tests/OpenIDCDiscoveryGrantTypesTest.java
@@ -1,0 +1,129 @@
+package io.vertx.tests;
+
+import io.vertx.core.http.HttpMethod;
+import io.vertx.core.http.HttpServer;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.auth.oauth2.OAuth2Options;
+import io.vertx.ext.auth.oauth2.impl.OAuth2AuthProviderImpl;
+import io.vertx.ext.auth.oauth2.providers.OpenIDConnectAuth;
+import io.vertx.ext.unit.Async;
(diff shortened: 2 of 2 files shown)


jnbdz:
Hi @pmlopes @vietj — gentle ping on this one when you have a moment. It's a one-line guard move in `OpenIDConnectAuth.discover` (only reset `supportedGrantTypes` when the server actually sends `grant_types_supported`), covered by `OpenIDCDiscoveryGrantTypesTest` against a local mock discovery server. Happy to adjust anything.

tsegismont:
For the record, there's another for the same issue https://github.com/eclipse-vertx/vertx-auth/pull/730

It takes a different approach (intersection of user-configured grant types with grant types advertised by the server).

PR #551, opened 2022-04-12
Fixed merging array-type attributes when only one of the two User objects has the attribute

dmironowicz:
See #550

PR #755, opened 2026-08-08
Do not wrap array attributes in a new array on User.merge

### LINKED ISSUE #550: User.merge puts array attribute in array
### Version

4.2.6

### Context
Merging two `User` objects, such that only one of them has an array attribute with a given name, results in the merged user having this attribute as array of arrays.

### Steps to reproduce

1. Create 1st User without any attributes
2. Create 2nd User with a single `JsonArray` attribute, e.g. `roles = ["write"]`
3. Call `merge` on the 1st user passing 2nd as parameter
5. Merged user has `roles = [ ["write"] ]`

jnbdz:
When merging two `User` objects where only the merged-in user has an array attribute for a given key, `User.merge` wrapped that array in a new array: `roles = ["write"]` became `roles = [["write"]]`.

The accumulate logic in `UserImpl.merge` handled every combination correctly (array+array concatenates, scalar+scalar becomes a two-element array, ...) except the case where the current user has no value for the key and the other user's value is an array — that branch added the whole array as a single element. This change keeps a copy of the other user's array instead, consistent with the copy the method already makes when the whole attributes object is absent.

Added coverage to `UserTest` for array attributes present on only one side of the merge (both directions).

Fixes #550

### DIFF
--- a/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/UserImpl.java
+++ b/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/UserImpl.java
@@ -139,7 +139,7 @@ public class UserImpl implements User, ClusterSerializable {
           Object rhsValue = otherAttrs.getValue(key);
           // accumulate
           if (lhsValue == null) {
-            attrs.put(key, rhsValue instanceof JsonArray ? new JsonArray().add(rhsValue) : rhsValue);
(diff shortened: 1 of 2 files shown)


tsegismont:
Duplicates https://github.com/eclipse-vertx/vertx-auth/pull/551

PR #756, opened 2026-08-09
Allow mapping authentication query columns to user attributes

### LINKED ISSUE #694: Return full user information on authentication query
#### Describe the feature

I have a specific improvement regarding this part:
https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-sql-client/src/main/java/io/vertx/ext/auth/sqlclient/impl/SqlAuthenticationImpl.java#L76-L82

The documentation states, that this library requires a table with a username column, and while this column name is hardcoded in the default query (https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-sql-client/src/main/java/io/vertx/ext/auth/sqlclient/SqlAuthenticationOptions.java#L35) which one can set, but in the code the column name is not mentioned, instead it it assumed that the password is the first string in a single database row.

The `User`object already accomodates for some more information to be put to it, as well as the `SqlAuthenticationOptions` hint at the idea of making this aspect at least somewhat configurable.

(By the way, I have to use this feature anyways, because Postgres expects the query placeholder to be `$1`.)

What I would like to have is that if one sets a different authentication query, one that returns more than a single value, for example `SELECT * ...`, and then including additional attributes could be implemented like this:

```
   // on SqlAuthenticationOptions
    public void addUserAttribute(String attributeName, Function<Row, Object> extractor) {
        userAttributeExtractors.put(attributeName, extractor);
    }
```
and usage:
```
SqlAuthenticationOptions().addUserAttribute("user_id", { row -> row.getUUID("user_id") } )
```

Maybe even this is possible:
```
SqlAuthenticationOptions().addUserAttribute("user_id", Row::.getUUID)
```

#### Use cases

Simply, when getting this User object from authentication, for the purpose of storing it in a token, or session, one saves haveing to query more information afterwards. Specifically, a `user_id` field might be something that users of this library would like to be returned as part of the `User`.

#### Contribution

I could possibly make a contribution, if the feature is deemed worth implementing.

jnbdz:
The SQL authentication provider discards every column of the authentication query except the password (`row.getString(0)`), so callers that need more user data right after login — typically the user id for a session or token — have to run a second query for the row they just fetched.

Following the guidance in #694, this adds a `SqlAuthentication.create(client, options, attributeMapper)` overload where the mapper is a `Function<Row, JsonObject>` living on the auth type (options stay data-only). The mapper receives the authenticated row and the returned JSON object is merged into the authenticated `User` attributes:

```java
AuthenticationProvider provider = SqlAuthentication.create(sqlClient,
  new SqlAuthenticationOptions()
    .setAuthenticationQuery("SELECT password, email FROM users WHERE username = ?"),
  row -> new JsonObject().put("email", row.getString("email")));
```

Behavior notes:
- the password is still expected in the first column, existing `create` overloads are unchanged
- a `null` mapper result is ignored
- a mapper that throws fails the authentication future
- the mapper only runs after the password has been verified

Includes a test in `MySQLTest` (the DDL gains a nullable `email` column), an example and documentation.

Fixes #694


vertx-auth-sql-client/src/main/java/io/vertx/ext/auth/sqlclient/SqlAuthentication.java
```diff
@@ -52,6 +55,24 @@ static SqlAuthentication create(SqlClient client, SqlAuthenticationOptions optio
...
+   * column is available to the given {@code attributeMapper}. The JSON object returned by the
+   * mapper is merged into the {@link io.vertx.ext.auth.User#attributes()} of the authenticated
+   * user.
+   *
+   * @param client          the JDBC client instance
+   * @param options         authentication options
+   * @param attributeMapper maps the authenticated row to extra user attributes, may return {@code null}
+   * @return the auth provider
+   */
+  static SqlAuthentication create(SqlClient client, SqlAuthenticationOptions options, Function<Row, JsonObject> attributeMapper) {
```
tsegismont, line 72: Please add `@GenIgnore(GenIgnore.PERMITTED_TYPE)`

vertx-auth-sql-client/src/main/java/io/vertx/ext/auth/sqlclient/SqlAuthentication.java
```diff
@@ -52,6 +55,24 @@ static SqlAuthentication create(SqlClient client, SqlAuthenticationOptions optio
     return new SqlAuthenticationImpl(client, options);
   }
 
+  /**
+   * Create a JDBC auth provider implementation that enriches the authenticated user with
```
tsegismont, line 59:
```suggestion
   * Create a SQL auth provider implementation that enriches the authenticated user with
```

Could you please also fix the other mentions of JDBC?

vertx-auth-sql-client/src/main/java/io/vertx/ext/auth/sqlclient/SqlAuthentication.java
```diff
@@ -52,6 +55,24 @@ static SqlAuthentication create(SqlClient client, SqlAuthenticationOptions optio
...
+   * attributes extracted from the authentication query row.
+   * <p>
+   * The authentication query is expected to return the password in the first column, any other
+   * column is available to the given {@code attributeMapper}. The JSON object returned by the
```
tsegismont, line 63: Can you please reword this part to make it clear that all columns returned by the authentication query are present? Including the hashed password in the first column. And that the user provided function should handle this information with care

vertx-auth-sql-client/src/main/java/io/vertx/ext/auth/sqlclient/impl/SqlAuthenticationImpl.java
```diff
@@ -77,6 +86,12 @@ public Future<User> authenticate(Credentials credentials) {
               User user = User.fromName(authInfo.getUsername());
               // metadata "amr"
               user.principal().put("amr", Collections.singletonList("pwd"));
+              if (attributeMapper != null) {
+                JsonObject attributes = attributeMapper.apply(row);
```
tsegismont, line 90: The mapper call should be wrapped with try/catch, log failures and return a failed future `Failure in authentication`

tsegismont (changes requested):
Thank you @jnbdz 

In addition to the comments inline, I think we should put the additional information into the user principal, not the attributes.

In vertx-auth, `principal()` holds **identity data** (who the user is) while `attributes()` holds
**authentication metadata** (decoded tokens, expiration timestamps, claims).

Every credential-based auth provider in the project (LDAP, htpasswd, htdigest, properties, OTP, WebAuthn, and SQL itself) puts all user data into `principal()` and leaves `attributes()` empty.

Only JWT and OAuth2 use `attributes()`, and exclusively for decoded token structures and timing metadata (`accessToken`, `idToken`, `exp`, `iat`, `nbf`, `rootClaim`, etc.).

Database columns like `email` or `display_name` describe the user's identity, they belong in `principal()`.

Merging them into `attributes()` is inconsistent with the rest of the codebase
and risks colliding with framework-managed keys that control `User.expired()`, `User.subject()`, `User.get()`, and authorization provider lookups.

PR #759, opened 2026-08-12
Rename misleading WebAuthn4J credential options methods

### LINKED ISSUE #698: Rename WebAuthn4J credential options methods to match W3C specification terminology
https://github.com/eclipse-vertx/vertx-auth/blob/f2fe95820b41db7da3afcc8a2028fedeb8b5ee8f/vertx-auth-webauthn4j/src/main/java/io/vertx/ext/auth/webauthn4j/WebAuthn4J.java#L64
https://github.com/eclipse-vertx/vertx-auth/blob/f2fe95820b41db7da3afcc8a2028fedeb8b5ee8f/vertx-auth-webauthn4j/src/main/java/io/vertx/ext/auth/webauthn4j/WebAuthn4J.java#L76

The names of these methods are misleading. 
When placed side by side, `createCredentialsOptions` appears to be method that create `CredentialsOptions` and `getCredentialsOptions` appears to just return `CredentialsOptions`.
However, they actually create `PublicKeyCredentialCreationOptions` and `PublicKeyCredentialGetOptions`, respectively.

Since introducing a new module `vertx-auth-webauthn4j` is a good chance to make a breaking change from `vertx-auth-webauthn`, I think these methods should be renamed to `createPublicKeyCredentialCreationOptions` and `createPublicKeyCredentialGetOptions`.

jnbdz:
As reported in #698, `createCredentialsOptions` and `getCredentialsOptions` read as if they create and return a `CredentialsOptions` object, while they actually create `PublicKeyCredentialCreationOptions` (for `navigator.credentials.create()`) and `PublicKeyCredentialRequestOptions` (for `navigator.credentials.get()`).

This introduces `createPublicKeyCredentialCreationOptions` and `createPublicKeyCredentialRequestOptions` as the primary methods, and turns the old names into `@Deprecated` default methods that delegate — the issue proposed a hard rename while the module was pre-release, but 5.x has since shipped with the old names (and `WebAuthn4JHandler` in vertx-web calls them), so a deprecation cycle seemed safer. If a hard break is preferred for the next major, the old defaults are trivial to drop.

Naming note: the issue suggested `createPublicKeyCredentialGetOptions` for the second method; this PR uses `RequestOptions` to match the spec dictionary name (`PublicKeyCredentialRequestOptions`) and WebAuthn4J's own class name — happy to adjust if the original suggestion is preferred.

Also fixes the javadoc of the request options method, which linked to the creation options dictionary of the spec, and updates docs/examples to the new names. Tests cover both the new names and the deprecated delegating aliases.

Fixes #698


tsegismont (approved):
LGTM, thank you

Issue #762, opened 2026-08-16
JWK.publicKey()/privateKey() return null for EC keys since 5.1


jnbdz:
### Version

5.1.x / master (regression from 5.0.x)

### Context

`JWK.publicKey()` and `JWK.privateKey()` return `null` for every EC key (`ES256`, `ES384`, `ES512`, `ES256K`), whether loaded from a JSON JWK, a PEM `PubSecKeyOptions` or a keystore. RSA, PS and EdDSA keys are unaffected.

Cause: a9d1ad4e ("Refactor JWT signing algorithm creation") replaced the stored `publicKey`/`privateKey` fields with `signingAlgorithm.unwrap()` and an `instanceof DigitalSigningAlgorithm` check. EC keys are wrapped by `JWK.wrapECAlgo(...)` (to convert JWS R+S signatures to/from ASN.1); that anonymous `SigningAlgorithm` does not override `unwrap()`, so the default `return this` stops the chain at the wrapper, which is not a `DigitalSigningAlgorithm`.

### Steps to reproduce

```java
JWK jwk = new JWK(new JsonObject()
  .put("kty", "EC").put("crv", "P-256")
  .put("x", "MKBCTNIcKUSDii11ySs3526iDZ8AiTo7Tu6KPAqv7D4")
  .put("y", "4Etl6SRW2YiLUrN5vfvVHuhp7x8PxltmWWlbbM4IFyM"));
assert jwk.publicKey() != null; // fails on 5.1.x, passed on 5.0.x
```

Same with `new JWK(new PubSecKeyOptions().setAlgorithm("ES256").setBuffer("-----BEGIN PUBLIC KEY-----..."))`.

Fix: override `unwrap()` in the EC wrapper to delegate to the wrapped algorithm. PR follows.

tsegismont:
Fixed by 4574a2392ca0ed1d4e41ed014ab3d78bc97379ea

PR #763, opened 2026-08-16
Expose key material of EC JWKs again

jnbdz:
See #762

Regression from a9d1ad4e: `JWK.publicKey()`/`privateKey()` now go through `signingAlgorithm.unwrap()`, but the anonymous EC wrapper created by `JWK.wrapECAlgo(...)` did not override `unwrap()`, so for every EC key (`ES256`/`ES384`/`ES512`/`ES256K`, from JSON JWK, PEM or keystore) the chain stopped at the wrapper and both accessors returned `null`. RSA/PS/EdDSA keys were unaffected because they are not wrapped.

Fix: the wrapper delegates `unwrap()` to the wrapped `DigitalSigningAlgorithm` (one method). Nothing else about the wrapper changes — signing/verification still convert between JWS R+S and ASN.1 as before.

Tests: `JWKTest` gains `publicECExposesPublicKey`, `privateECExposesKeyPair`, `pemECExposesKeys` (all failed before the fix) and `pemRSAExposesPublicKey` as a control. vertx-auth-common, vertx-auth-jwt and vertx-auth-oauth2 suites are green locally.

### DIFF
--- a/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java
+++ b/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/jose/JWK.java
@@ -325,6 +325,11 @@ public final class JWK {
     // for all JWS, it may not be true for COSE keys
     return new SigningAlgorithm() {
       @Override
+      public SigningAlgorithm unwrap() {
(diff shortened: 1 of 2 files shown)


tsegismont (approved):
LGTM, thanks

